PR #38, opened 2025-03-24
Allow to disable cookies secure flag when needed

### LINKED ISSUE #13: Invalid CSRF token
Hi,

I followed your instructions for a docker installation on my Ubuntu Server 24.04.2 LTS.
The three containers are up and running. But when I logg into die WebUI I get the Error "Invalid or expired session" and when I want to save
new settings I've got: "Invalid CSRF token".

Here is a snap from the container log:

timestamp=2025-02-28T10:10:50.653Z level=error message="error validating session" caller=/home/runner/work/libredesk/libredesk/cmd/middlewares.go:51 sc=libredesk error=null 
timestamp=2025-02-28T10:10:50.676Z level=error message="csrf token mismatch" caller=/home/runner/work/libredesk/libredesk/cmd/middlewares.go:81 sc=libredesk cookie_token= header_token= 

thank you for your support

jleroy:
When Libredesk is accessed through non-HTTPS connections (typically on development environments), both CSRF and session cookies are still sent with a `Secure` flag, which means they're ignored by the browser, resulting in `Invalid CSRF token`and `Invalid or expired session` errors in the app.

This PR add an new `disable_secure_cookies` option which allow to disable the secure flag being set on both CSRF an session cookies when enabled.

Please note that:
 - I’m not a Go specialist and Libredesk auth code is a bit complex, so maybe there’s a better way to do that than using Auth.cfg?
 - I haven’t modified `simpleSessGetCookieCB()` and `simpleSessSetCookieCB()` as I don’t think this is necessary? 

Fixes #13.

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -46,7 +46,8 @@ type Provider struct {
 
 // Config stores multiple OIDC provider configurations
 type Config struct {
-	Providers []Provider
+	Providers     []Provider
+	SecureCookies bool
 }
 
 // Auth is the auth service it manages OIDC authentication and sessions
@@ -92,7 +93,7 @@ func New(cfg Config, rd *redis.Client, logger *logf.Logger) (*Auth, error) {
 		Cookie: simplesessions.CookieOptions{
 			Name:       "libredesk_session",
 			IsHTTPOnly: true,
-			IsSecure:   true,
+			IsSecure:   cfg.SecureCookies,
 			MaxAge:     time.Hour * 9,
 		},
 	})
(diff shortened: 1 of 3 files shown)


yc01:
I would recommend not changing this but using something like [mkcert](https://github.com/FiloSottile/mkcert) to use https with localhost

jleroy:
Using SSL certificates requires to setup a webserver like Nginx in front of Libredesk which is a bit too much for just testing it or running it as a developer.
Not everyone uses Docker or have a full development environment on its computer.

internal/auth/auth.go
```diff
@@ -46,7 +46,8 @@ type Provider struct {
 
 // Config stores multiple OIDC provider configurations
```
abhinavxd, line 47: Can update comment to 
```
// Config holds OIDC providers and cookie security settings.
```
to reflect purpose.

abhinavxd (changes requested):

abhinavxd:
LGTM

jleroy:
@abhinavxd Done.

abhinavxd:
Thank you @jleroy

PR #67, opened 2025-04-15
Add new actions "Add tags" and "Remove tags" for macros and automations

jleroy:
This PR add two new actions for macros and automations: "Add tags" and "Remove tags". 

Unlike "Set tags" who replace all existing tags, these actions allow to add/remove tags without touching existing ones. This is especially useful if you’re using tags for easily tracking SLA breach.

It also add tracking of tags removal in conversation activity.


abhinavxd:
Great addition, thanks @jleroy

PR #69, opened 2025-04-15
Add new automations rules filters "Hours since first reply" and "Hours since last reply"

jleroy:
This PR adds two new filters for conversations in automations:  `Hours since first reply` and `Hours since last reply`.

This adds a new `last_reply_at` field in the `conversations` table.
Contrary to the existing `last_message_at` field,  `last_reply_at` is updated **only** when a non-automated reply is sent (this excludes automated outgoing messages, conversation activities, privates notes...).

I’ve also removed `Hours since created` and `Hours since resolved` filters from `newConversationFilters` as these fields should always be NULL or equal to 0 when a conversation is created? Tell me if you want me to rollback this change.


jleroy:
> Yes makes sense these can be removed, can you re-base and fix the conflicts?

Done!

Issue #86, opened 2025-05-08
Feature - Add support for Next Response Time SLA metric


abhinavxd:
Libredesk currently supports only First Response Time and Resolution Time SLA metrics.
This issue tracks adding support for a third metric: Next Response Time, which measures the time taken for subsequent replies after the first response.

PR #95, opened 2025-05-22
Feature : Next response SLA metric

### LINKED ISSUE #70: SLA Alert triggers even if tickets are closed
Hi,

I just received many many e-mails regarding SLA deadline breaches from already closed tickets.

The tickets were just test tickets, so they were closed without any reponse, but it seems like the job that checks for SLA breaches doesnt care if a ticket is closed or not. I think this behaviour is not intended and the job should only run for active tickets.

Maybe this also requires a new attribute in the statuses section (beides the name and created_at a is_active). So when creating new statuses you can select if this ticket should count as active or not.

abhinavxd:
Ref #86

- Introduced new table `sla_events` to track next response metric details. A new table was necessary (in addition to the existing `applied_sla`) because next response events can occur multiple times.
- A reply from a contact creates an `sla_events` entry. The next agent reply sets the `met_at` timestamp for the latest entry.
- Notifications for next response time metric.
- SLA Form: Added support to set the metric type for each SLA notification, providing more control over notifications.

image

- SLA Form: Made first response time, resolution time, and next response time optional, with at least one required.

Other fixes
- Missed websocket updates for SLA timestamps.
- SLA badge re-render not happening when conversation timestamps are updated.
- Fixes #70


abhinavxd:
TODO for me
- Keep new i18n keys consistent, can use existing instead of introducing new keyts.
- Query struct query names in `sla` package can be named better as there are 3 tables involved, SLA policy, Applied SLA, SLA events.

PR #123, opened 2025-07-24
Feat/bulk import agent

venkataprasathbaskaran:
Screenshot 2025-07-24 at 10 25 13 PM
Screenshot 2025-07-24 at 10 25 16 PM
Screenshot 2025-07-24 at 10 25 22 PM

Hi @abhinavxd 

I saw the feature request of bulk csv import of agent. So wanted to give it a try and contribute to FOSS. I have implement pretty basic stuff of import via API. 

Everything happens in API call itself for now. ideally this will scale when used for large files. Where we might need to implement a background worker for process file and write report of import and have resislency in place to tackle worker craches. For start i though we can implement this basic import where they can import 100 agent in a single call. 

I am raising this PR with screenshot of the API and csv used. You can give feedback and suggestions so i can improve on this. 

Thanks!

### DIFF
--- a/cmd/handlers.go
+++ b/cmd/handlers.go
@@ -115,6 +115,7 @@ func initHandlers(g *fastglue.Fastglue, hub *ws.Hub) {
 	g.POST("/api/v1/agents/reset-password", tryAuth(handleResetPassword))
 	g.POST("/api/v1/agents/set-password", tryAuth(handleSetPassword))
 	g.POST("/api/v1/agents/import", perm(handleBulkImportAgents, "users:manage"))
+	g.GET("/api/v1/agents/import/status", perm(statusHandler, "users:manage"))
 
 	// Contacts.
 	g.GET("/api/v1/contacts", perm(handleGetContacts, "contacts:read_all"))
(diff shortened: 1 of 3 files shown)


abhinavxd:
Thank you for the PR! I will check it over the weekend.

abhinavxd:
Thanks for the PR!

I propose creating an importer package that’s initialized once on `App` struct, allowing uploads of more than 100 records without hitting the HTTP request timeout.

It accepts imports by namespace so status can be polled later.

Something like this- 
```go
package importer


type JobStatus struct {
    Running   bool   
    Logs      []string  
    Total     int       
    Success   int    
    Errors    int      
    StartedAt time.Time
    EndedAt   time.Time
}

type Importer struct {
    jobs map[string]*JobStatus
}

func NewImporter() *Importer {
    i := &Importer{
        jobs: make(map[string]*JobStatus),
    }
    // could start a new goroutine here that cleans up jobs every hour
    go i.cleanUp()
    return i
}

func (i *Importer) Submit(namespace string, fn func() error) error {
    // ...
    fn()
    // ...
}
```

On the front end, replace the import button with a common Importer.vue component that opens a dialog, lets the user select a file, use clicks import. We can poll:
```
GET - /api/v1/agents/import/status 
```
for logs and progress. This lets us reuse the importer across the app.

Let me know what you think!

PR #171, opened 2025-11-19
Fix/reply draft persistence

csr4422:
https://github.com/user-attachments/assets/a871bdb9-6327-48ec-9bd2-64703c8a3573

# Conversation Draft Persistence

A new draft-saving system has been introduced to improve the reply experience by persisting drafts to the backend database.

## Features

- Stores conversation drafts using Pinia `draftStore`
- **Backend Persistence**: Drafts saved to PostgreSQL database for cross-device/browser access
- Automatically restores drafts when returning to a conversation
- Debounced autosave (500ms) for smooth performance and reduced server load
- Immediate draft clearing after successful message send
- **Access Control**: Enforces conversation access permissions for all draft operations
- **Single Source of Truth**: All draft data stored exclusively in database

### Backend
- Created `conversation_drafts` database table with unique constraint per user/conversation
- Implemented CRUD operations for drafts (upsert, read, delete)
- Added authenticated API endpoints:
  - `POST /api/v1/conversations/{uuid}/draft` - Create/update draft
  - `GET /api/v1/conversations/{uuid}/draft` - Retrieve draft
  - `DELETE /api/v1/conversations/{uuid}/draft` - Delete draft
- Added conversation access control and content validation
- Upsert support using `ON CONFLICT DO UPDATE` for atomic operations

### Frontend
- Integrated backend API with Pinia store for draft management
- Made draft loading async to fetch from backend database
- Drafts persist across devices and browser sessions
- Real-time sync on every change (debounced 500ms)

### Backend
```
cmd/draft.go
cmd/handlers.go
internal/conversation/conversation.go
internal/conversation/models/models.go
internal/conversation/queries.sql
migrations/xxx_add_conversation_drafts.sql
```

### Frontend
```
src/stores/draftStore.js
src/composables/useDraftManager.js
src/api/index.js
ReplyBox.vue
TextEditor.vue
```

### DIFF
--- a/frontend/src/composables/useDraftManager.js
+++ b/frontend/src/composables/useDraftManager.js
@@ -57,7 +57,6 @@ export function useDraftManager (key, uploadedFiles = null) {
   const conversationStore = useConversationStore()
   const htmlContent = ref('')
   const textContent = ref('')
-  const draftMeta = ref({})
   const isLoading = ref(false)
   const isDirty = ref(false)
   const skipNextSave = ref(false)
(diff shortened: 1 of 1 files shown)


frontend/src/features/conversation/ReplyBox.vue
```diff
@@ -151,6 +152,15 @@ const { uploadingFiles, handleFileUpload, handleFileDelete, mediaFiles, clearMed
     linkedModel: 'messages'
   })
 
+// Setup draft management composable
+const currentDraftKey = computed(() => conversationStore.current?.uuid || null)
+const {
+  htmlContent: localHtmlContent,
+  textContent: localTextContent,
```
abhinavxd: If you don't alias it.


Then most of the code remains unchanged. Previously var was named `htmlContent`

OfficialCRUGG:
Does it really make sense to store drafts in local storage?
I could for example imagine a scenario where you start drafting a reply to a ticket on your laptop on the go, and then want to finish it at your PC at home.
Or even with a phone, once support is better for that.

All in all, wouldn't it make way more sense to store drafts per account in the DB and not per device?

PR #183, opened 2025-11-29
fix: macro list now updates reactively on create/delete

csr4422:
## Description
Fixes the issue where Inbox component does not reactively update when a new macro is added or deleted.

## Changes
- Modified `loadMacros` to accept force parameter for reloading from server
- Call `loadMacros(true)` after create/update/delete operations to refresh macro list

## Testing
https://github.com/user-attachments/assets/1fe0f654-2550-4e6a-a9d1-23f33bccdf45
- Create new macro - appears in list without reload
- Delete macro - removed from list without reload

### DIFF
--- a/frontend/src/views/admin/macros/EditMacro.vue
+++ b/frontend/src/views/admin/macros/EditMacro.vue
@@ -16,12 +16,14 @@ import MacroForm from '@/features/admin/macros/MacroForm.vue'
 import { CustomBreadcrumb } from '@/components/ui/breadcrumb'
 import { useI18n } from 'vue-i18n'
 import { Spinner } from '@/components/ui/spinner'
+import { useMacroStore } from '@/stores/macro'
 
 const macro = ref({})
 const { t } = useI18n()
 const isLoading = ref(false)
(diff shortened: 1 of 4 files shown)


frontend/src/views/admin/macros/EditMacro.vue
```diff
@@ -36,11 +40,16 @@ const updateMacro = async (payload) => {
...
+    
+    // Reload macros from server
+    await macroStore.loadMacros(true)
+    
     emitter.emit(EMITTER_EVENTS.SHOW_TOAST, {
       description: t('globals.messages.updatedSuccessfully', {
         name: t('globals.terms.macro')
       })
     })
+        router.push({ name: 'macro-list' })
```
abhinavxd: Remove this.

abhinavxd (changes requested):

frontend/src/views/admin/macros/EditMacro.vue
csr4422: yes

PR #256, opened 2026-03-02
fix(smtp): try to authenticate with the UPN from OAuth

lucaasmth:
## Summary

This pull request improves the handling of OAuth authentication for Microsoft SMTP by ensuring the XOAUTH2 username matches the User Principal Name (UPN) in the JWT access token, which is required for successful authentication. The changes introduce a helper function to extract the UPN from the token and update the authentication logic accordingly.

In my case I had two different emails on the Microsoft account and I couldn't authenticate to the SMTP server with the first, which is the one present in the email field of the JWT access token.

## Changes

* Added a new helper function `extractUPNFromToken` in `internal/inbox/channel/email/smtp.go` to extract the UPN (or fallback to `unique_name`) from a JWT access token, ensuring proper username selection for Microsoft SMTP.
* Updated the `NewSmtpPool` function to use the extracted UPN as the XOAUTH2 username if available, otherwise defaulting to the configured username, improving compatibility with Microsoft SMTP requirements.

### DIFF
--- a/internal/inbox/channel/email/smtp.go
+++ b/internal/inbox/channel/email/smtp.go
@@ -37,8 +37,12 @@ func NewSmtpPool(configs []imodels.SMTPConfig, oauth *imodels.OAuthConfig) ([]*s
 
 		// Check if OAuth authentication should be used
 		if oauth != nil && oauth.AccessToken != "" {
+			username := cfg.Username
+			if upn := extractUPNFromToken(oauth.AccessToken); upn != "" {
+				username = upn
+			}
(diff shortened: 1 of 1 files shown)


abhinavxd:
hey thanks for the PR!

This needs to be also done for imap I believe not just SMTP.

abhinavxd:
this doesn't compile, will create a new PR for this

abhinavxd:
I've pushed a fix here - https://github.com/abhinavxd/libredesk/pull/536

If possible can you try if this works? Thank you

PR #260, opened 2026-03-05
Fix: update contact names on upsert instead of discarding them

josephsellers:
## Summary

When a contact already exists, the `insert-contact` query currently only updates `updated_at` on conflict. This means `first_name` and `last_name` values provided via the API (e.g. from form submissions or email workers) are silently discarded for existing contacts.

This one-line change updates the ON CONFLICT clause to set `first_name` and `last_name` from the incoming values, using `COALESCE(NULLIF(EXCLUDED.x, ''), users.x)` so that:
- Non-empty incoming names replace existing ones
- Empty strings are ignored, preserving existing names
- `updated_at` is still touched as before

## Changes

- `internal/user/queries.sql`: Modified `insert-contact` ON CONFLICT clause

## Test plan

- Create a contact via API with a name
- Create the same contact again (same email) with a different name
- Verify the name is updated
- Create the same contact again with an empty name
- Verify the existing name is preserved

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved contact creation workflow by correctly matching and reusing existing contacts via email when enrichment identifiers are unavailable.
  * Enhanced data integrity by preventing empty field values from inadvertently overwriting preserved contact information during updates.

### DIFF
--- a/internal/user/contact.go
+++ b/internal/user/contact.go
@@ -51,20 +51,22 @@ func (u *Manager) CreateContact(user *models.User) error {
 	}
 
 	if user.Email.Valid && user.Email.String != "" {
-		// Reuse any existing contact with this email, preferring one with ext_id if multiple exist.
+		// An ext_id contact owns this email - reuse it; the no-ext-id upsert below can't match it and would insert a duplicate.
 		existing, err := u.GetContactByEmail(user.Email.String)
-		if err == nil {
+		if err == nil && existing.ExternalUserID.String != "" {
 			user.ID = existing.ID
 			return nil
 		}
 
 		// Other error than not found - fail.
-		if envErr, ok := err.(envelope.Error); !ok || envErr.ErrorType != envelope.NotFoundError {
-			return err
+		if err != nil {
+			if envErr, ok := err.(envelope.Error); !ok || envErr.ErrorType != envelope.NotFoundError {
+				return err
+			}
 		}
 	}
 
-	// No ext_id and no existing contact with email - create new.
+	// No ext_id contact for this email - insert new, or update the existing no-ext-id contact's name.
 	if err := u.q.InsertContactNoExtID.QueryRow(user.Email, user.FirstName, user.LastName, password, user.AvatarURL).Scan(&user.ID); err != nil {
 		u.lo.Error("error inserting contact", "error", err)
 		return fmt.Errorf("insert contact: %w", err)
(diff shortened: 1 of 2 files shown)


abhinavxd:
Thanks for the PR!

This change would allow agents without the "Edit contact details" permission to update a contact’s name while creating a new conversation for an existing contact.

This contradicts the permission model. In the latest build, the first name and last name fields are disabled in the form.

It’s better to keep the current behavior and require agents to update contact details from the contact view page.

josephsellers:
Thanks for the review, and apologies for the slow turnaround.

I want to make sure I understand the concern, and propose a narrower fix.

**Use case:** Our integrations create conversations via the public API rather than the agent UI: a Cloudflare Worker for inbound email, and a WordPress (Gravity Forms) plugin posting support requests. In both cases the first/last name come from authoritative sources (email `From` header, form fields) and `initiator` is always `contact`. For existing contacts, the names are silently dropped today, which means name corrections from these sources never reach LibreDesk.

**On the permission concern:** I agree that letting an agent overwrite a contact's name through the new-conversation form would contradict the permission model. However, since the form disables the first/last name fields for existing contacts in the latest build, the values posted with `initiator: agent` already match what's on file, so the upsert is a no-op for that path in practice.

**Proposed narrower fix:** scope the name update to non-agent flows. The smallest version is a one-liner in `handleCreateConversation`:

```go
if req.Initiator == umodels.UserTypeAgent {
    contact.FirstName = ""
    contact.LastName = ""
}
```

The existing `NULLIF(EXCLUDED.first_name, '')` then preserves the stored name on the agent path, while contact-initiated and IMAP flows continue to update names as the SQL change intends. This keeps the agent UI path free of any permission-bypass risk.

Happy to push this update if it sounds reasonable, or split into two SQL queries (`insert-contact` vs `upsert-contact`) if you'd prefer the intent to be visible at the query layer instead.

abhinavxd:
Hey, sorry I missed this.

Revisited this, you are right.

Upsert is keyed on (email, type), so only someone controlling that email can trigger the update, which is effectively the contact updating their own name.

There's a conflict once you've fixed it, we can merge this.

josephsellers:
Rebased onto current `main` and resolved the conflict. The contact model changed since this PR was opened (`external_user_id`), so the fix now targets `insert-contact-without-external-id` — `insert-contact-with-external-id` already updates names on conflict. Same COALESCE(NULLIF(...)) approach so empty values don't clobber existing names. Ready when you are, thanks!

abhinavxd:
I will add to the next release 2.4.0, releasing this weekend.

Again thank you!

abhinavxd:
Thank you so much!

PR #261, opened 2026-03-05
Add API endpoint to delete private notes

josephsellers:
## Summary

Adds `DELETE /api/v1/conversations/{cuuid}/messages/{uuid}` for removing private notes via the API. Only messages with `private=true` can be deleted; the SQL WHERE clause ensures sent messages are protected. Returns 404 if the message doesn't exist or isn't private.

This is useful for API consumers that create private notes programmatically and need to clean them up (e.g. replacing an outdated note before adding a new one).

## Changes

- `cmd/messages_delete.go`: New handler (`handleDeleteMessage`) with conversation access enforcement
- `internal/conversation/message_delete.go`: New `DeleteMessage` method on the conversation manager
- `cmd/handlers.go`: Route registration (DELETE, `messages:write` permission)
- `internal/conversation/conversation.go`: `DeleteMessage` prepared statement in query struct
- `internal/conversation/queries.sql`: `delete-message` query

## Test plan

- Create a private note via `POST /conversations/{uuid}/messages` with `private: true`
- Delete it via `DELETE /conversations/{cuuid}/messages/{uuid}`
- Verify it returns success and the note is gone
- Attempt to delete a non-private (sent) message
- Verify it returns 404 (protected by SQL constraint)
- Attempt to delete a message from a conversation the agent doesn't have access to
- Verify it returns a permission error

## Summary by CodeRabbit

* **New Features**
  * Added “Delete note” for private message bubbles via a dropdown action, with a confirmation dialog.
  * Added a client API and conversation store action to soft-delete a private note by conversation/message UUID and update the UI immediately.
* **Localization**
  * Added English strings for “Delete note”, the permanent-deletion confirmation, and successful deletion messaging.
* **Backend**
  * Added a protected DELETE endpoint for deleting only private messages in a conversation (soft-delete), returning not found when the note can’t be deleted.


abhinavxd:
Hey, thanks for the PR! I will merge this as soon as I merge - https://github.com/abhinavxd/libredesk/pull/114

#114 PR adds the live chat widget, which required me to refactor and move many parts of the project. Since everything would need to be retested anyway, I ended up bundling in a lot of other fixes as well to avoid multiple rounds of QA.

abhinavxd:
Hey there seems to be conflict due to recent live chat branch merge

josephsellers:
Rebased onto current `main` and resolved the live-chat conflict.

Two adjustments for v2:
- **Backend:** `main` now has a generic `delete-message` query/stmt (used internally by continuity), so to avoid a collision this uses a dedicated **`delete-private-message`** query + `DeletePrivateMessage` stmt/method, still guarded to `private = true`.
- **Frontend:** the context menu is re-wired against the refactored `MessageBubble.vue` (wrapped in `ContextMenu` with `as-child`, trigger disabled for non-private messages). Much smaller diff this time.

Builds clean (Go + `pnpm build:main`). Ready for another look, thanks!

josephsellers:
Good catch from the CodeRabbit review on the IDOR — fixed in `406a2941`. `delete-private-message` is now scoped to the conversation (`conversation_id = (SELECT id FROM conversations WHERE uuid = $2)`), and `DeletePrivateMessage` + the handler pass both the conversation uuid and the message uuid. So a note can only be deleted via the conversation it actually belongs to, even though the handler already gates access to that conversation. Verified the cross-conversation case now returns 404 and leaves the note intact.

PR #262, opened 2026-03-06
Add configurable endpoint URL and model for AI Assist

josephsellers:
## Summary

The AI Assist feature currently hardcodes `api.openai.com` and `gpt-4o-mini`, with only the API key editable via an inline popup. This makes it unusable with alternative providers or self-hosted models.

This PR adds admin-configurable **endpoint URL** and **model** fields alongside the existing API key, so any OpenAI-compatible API can be used — including self-hosted models (vLLM, Ollama, llama.cpp), alternative providers (Groq, Together AI, Mistral), or proxied endpoints.

## Changes

**Backend** (4 files):
- `internal/ai/openai.go` — Configurable `endpointURL` and `model` fields with sensible defaults (`api.openai.com`, `gpt-4o-mini`). HTTP timeout increased from 10s to 30s (self-hosted models and some providers can exceed 10s for completions).
- `internal/ai/ai.go` — New `GetProvider()` method for reading config (sanitized, no decrypted key). `UpdateProvider()` extended with `endpointURL`/`model` params and server-side URL validation. Read-modify-write pattern preserves existing encrypted API key when only changing endpoint or model.
- `internal/ai/queries.sql` — Replace `set-openai-key` with `update-provider-config` (full JSONB write with `updated_at`).
- `cmd/ai.go` + `cmd/handlers.go` — New `GET /api/v1/ai/provider` endpoint (requires `ai:manage`), extended PUT payload with optional `endpoint_url` and `model` fields.

**Frontend** (6 files):
- New admin settings page at **Admin → Integrations → AI Assist** (`AISettings.vue`, `AISettingsForm.vue`, `formSchema.js`).
- Endpoint URL, model, and API key fields with Zod validation (URL format when non-empty).
- Green "API key is configured" indicator when a key exists.
- Link from the existing inline API key dialog in `ReplyBox.vue` to the new settings page ("For advanced setup").
- `navigation.js` — AI Assist item added under Integrations.
- `router/index.js` — `/admin/ai` route.

**i18n** (1 file):
- 9 new keys in `en.json` for the settings UI.

## Backwards compatibility

- **No database migration needed.** The existing JSONB `config` column on `ai_providers` absorbs the new fields (`endpoint_url`, `model`) naturally. Go code defaults handle missing fields on existing installs — behaviour is identical to current upstream when fields are empty.
- **No breaking API change.** The PUT endpoint accepts `endpoint_url` and `model` as optional fields. Existing API callers sending only `{provider, api_key}` continue to work unchanged.
- **Existing inline API key dialog still works.** The new admin page is an addition, not a replacement.

## Screenshots

The new admin settings page follows the existing LibreDesk admin page patterns (AdminPageWithHelp layout, vee-validate + Zod, same component library).

## Test plan

- [ ] Fresh install: verify AI Assist works with default OpenAI endpoint after entering only an API key
- [ ] Existing install: verify upgrade with no migration, existing API key preserved
- [ ] Set custom endpoint URL and model, verify completions use the configured endpoint
- [ ] Leave endpoint/model empty, verify defaults (api.openai.com, gpt-4o-mini) are used
- [ ] Update only endpoint/model without re-entering API key, verify key is preserved
- [ ] Enter invalid URL (no http/https prefix), verify server-side validation rejects it
- [ ] Verify "advanced setup" link appears in ReplyBox inline dialog
- [ ] Verify AI Assist navigation item appears under Admin → Integrations

### DIFF
--- a/cmd/ai.go
+++ b/cmd/ai.go
@@ -11,8 +11,10 @@ type aiCompletionReq struct {
 }
 
 type providerUpdateReq struct {
-	Provider string `json:"provider"`
-	APIKey   string `json:"api_key"`
+	Provider    string `json:"provider"`
+	APIKey      string `json:"api_key"`
+	EndpointURL string `json:"endpoint_url"`
+	Model       string `json:"model"`
 }
 
 // handleAICompletion handles AI completion requests
@@ -45,6 +47,18 @@ func handleGetAIPrompts(r *fastglue.Request) error {
 	return r.SendEnvelope(resp)
 }
 
+// handleGetAIProvider returns the current AI provider config (sanitized).
+func handleGetAIProvider(r *fastglue.Request) error {
+	var (
+		app = r.Context.(*App)
+	)
+	resp, err := app.ai.GetProvider()
+	if err != nil {
+		return sendErrorEnvelope(r, err)
+	}
+	return r.SendEnvelope(resp)
+}
+
 // handleUpdateAIProvider updates the AI provider
 func handleUpdateAIProvider(r *fastglue.Request) error {
 	var (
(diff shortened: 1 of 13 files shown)


sumanthneerumalla:
requesting that this be reopened, if this feature is not already supported

Issue #259, opened 2026-03-04
My work on LibreDesk over the last month


mageaustralia:
Hi @abhinavxd ,

First off, thanks for releasing LibreDesk for everyone to use and modify. It's actually a really great system. When I first tested it, I made a few modifications to see how feasible for our needs it was.

I did try and submit some of those changes as PR's but our project started moving quickly and I've been pumping through many many changes per day that it was hard to submit all as PR's and align our business goals with those goals of the LibreDesk project overall. 

In any case, I wanted to bring our branch to your attention as I think there's a lot there that could be valuable to the LibreDesk community. I have no intention of running my own fork and I'd rather see much or all of this brought back into LibreDesk so we're not running our own fork in isolation but I do realise not all of it will fit with your goals.

My branch is here
https://github.com/mageaustralia/libredesk/tree/v1.0.3-plus-enhancements

And it contains all the following:

  **AI & Knowledge**
                                                                                                                                                                                                          
  - OpenRouter AI provider — 100+ models via OpenRouter, encrypted API key storage
  - RAG AI assistant — "Generate Response" button, knowledge sources (webpages, macros), system prompt config
  - Multimodal AI — Conversation images extracted, resized to 500x500, included in AI prompts
  - External search API — Pluggable search integration for RAG pipeline
  - AI HTML sanitization — Prevents prompt injection in AI-generated content

  **Ecommerce Integration**

  - Maho Commerce / Magento 1 provider — OAuth2 auth, customer & order lookup
  - Multi-stage context gathering — Auto-fetches customer + recent orders, scans conversation for order numbers
  - "+ Orders" button in reply box for AI responses with ecommerce context
  - Ecommerce settings page at /admin/ecommerce

  **Email & IMAP**

  - Per-inbox email signatures with dynamic placeholders, From switcher
  - Auto-assign on reply — Per-inbox toggle to assign conversation to replying agent
  - Agent name in From header — Emails show agent's name, not just inbox name
  - IMAP/SMTP connection testing with debug logs
  - SMTP test for notification settings
  - Multi-folder IMAP polling — Comma-separated mailbox folders
  - Email HTML sanitization — Clean incoming email HTML
  - DMARC rewrite detection — Detects real sender from forwarded emails (X-Google-Original-From)
  - Contact form field parsing — Extracts real customer name/email from HTML form bodies
  - Inline CID fix — Non-image attachments (PDF) render as download links instead of broken images
[...]

PR #286, opened 2026-04-20
Add bulk actions on conversations (assign, status, priority)

mageaustralia:
Adds bulk actions on conversations, per the item you mentioned in #259. Focused PR as you suggested there.

## What

Per-row checkboxes on the conversation list. When at least one conversation is selected, the filters bar swaps for a small toolbar with three dropdowns: Assign (agent or team), Status, Priority. Shift-click selects a range. The X on the right clears.

## How

Frontend only. The bulk operations call the existing per-conversation API endpoints (`updateAssignee`, `updateConversationStatus`, `updateConversationPriority`) in parallel via `Promise.allSettled`. No new backend handler.

That keeps the diff small. Trade-offs worth flagging:

* 50 selected items = 50 concurrent HTTP requests. Wall-clock for a typical bulk op is sub-second on a fast network, but it's not ideal at scale.
* No transactionality. If some calls fail, you get a toast like "Updated 47, failed 3 of 50" and the per-uuid errors are logged to console.

## Permissions

Each dropdown is gated on the corresponding existing permission:

* Assign Agents: \`conversations:update_user_assignee\`
* Assign Teams: \`conversations:update_team_assignee\`
* Status: \`conversations:update_status\`
* Priority: \`conversations:update_priority\`

A user with none of these still sees the count and clear button (since they could already select rows).

## i18n

8 new keys under \`conversation.bulkActions.*\` in \`i18n/en.json\`. Reused \`globals.terms.{agent,team,status,priority,error}\` for the singular and plural labels, so the Crowdin export is small. Other locales fall back to \`en\` until translated.

## Testing

Tested manually against a cloned production dataset (~5k conversations, ~80k contacts). Verified:

* Single selection, multi-selection, shift-click range
* Bulk assign agent, bulk assign team
* Bulk change status, bulk change priority
* Clear selection, including after a partial-failure run
* Toolbar shows the correctly localized count (singular vs plural)
* Permission gating verified by removing perms from a test role


abhinavxd:
Hey thanks for the PR! This is amazing and is exactly what we needed at work.

I tried it locally few suggestions - 

- Place the checkbox on top of the avatar and show it on hover. This saves space and removes the need for a permanently visible checkbox on the left.
image

<br>
<br>



- The top options could be just icons to save space. 
image

<br>


We can also allow changing the assigned team separately using the SelectCombobox component, since it supports search filter.
We could also allow setting tags.

Overall: assign team, assign agent, set tags, and status - all with icons to save space, with tooltips on each icon.
We can drop priority; not having it will space space.

abhinavxd:
We needed this feature soon, so I ended up making the changes.

PR #289, opened 2026-04-20
Inline images, lightbox, and attachment UX

mageaustralia:
## Summary
- **Editor**: paste/drop to upload images with drag-resize handles + size toolbar; resize handle uses pointer events so mouse, touch, and pen all work
- **Viewer**: image lightbox with zoom/pan, keyboard nav, and download. File attachments get type-specific icons; PDFs preview inline
- **Outgoing email**: inline images embed as `multipart/related` CID parts so recipients see them even if our `/uploads/` URLs later expire or the server is unreachable. Quoted-thread images from prior replies are picked up automatically. Cross-conversation references are rejected; orphan uploads (just pasted, not yet linked to a message) are allowed since they can't be used to exfiltrate from another conversation
- Tightens HTML safety in `replaceCIDInContent` and scopes the `/uploads/<uuid>` rewrite regex to `` attributes only

## Test plan
- [ ] Paste an image into the reply editor — uploads, inserts, drag-resize works, size toolbar appears on selection
- [ ] Drop an image file onto the editor — same flow as paste
- [ ] Drop a non-image file onto the editor — emits `filesDropped` for the parent attachment list
- [ ] Click an image in a received message — lightbox opens; zoom in/out, arrow keys move between images, download works
- [ ] Send a reply with a pasted image — inspect raw email source: `` and a matching inline `multipart/related` part
- [ ] Reply within a thread containing prior inline images — quoted-thread images embed as CID too
- [ ] Send a reply where the body references `/uploads/<uuid>` from a different conversation — rejected, src stripped, server log shows the warning
- [ ] PDF attachment renders the file-icon card; click the eye icon to preview inline


abhinavxd:
Hey thanks for the PR!
This along with bulk actions (for which you have a PR) is exactly what we needed at work!

Thank you so much!

PR #290, opened 2026-04-21
Edit and soft-delete private notes

mageaustralia:
> **Note:** I only spotted #261 after this work was done — was searching for open issues requesting note edit/delete and missed the PR. Different scope and design choice from #261 (which is API-only hard-delete). Happy to align with that PR's author if maintainers prefer one approach. Posting this in case it's useful as a starting point.

## Summary
- `PUT` and `DELETE /conversations/:cuuid/messages/:uuid/note` (perm `messages:write`)
- Authorisation: only the original author can edit/delete; the SQL also guards on `private = true AND type = 'outgoing'` so a public reply can't be touched accidentally
- **Edit** replaces the body in place via a textarea on the bubble (deliberately simple — Tiptap-for-notes can be a follow-up)
- **Delete** is a soft-delete: row stays, content becomes a tombstone naming the actor (`This note was deleted by <Name>`), and `meta.deleted = true` so the UI hides further edit/delete controls. Audit trail is preserved.

## Reactivity fixes that the new flow exposed
- `MessageCache.updateMessage` was `Object.assign`-ing in place, so consumers holding the message ref never saw the change. Now swaps the array entry to a new object so Vue picks up the identity change.
- `conversationMessages` computed didn't depend on `messages.version`, so in-place cache updates couldn't trigger a re-render. Now reads the version.
- `<Letter>` in `MessageBubble` doesn't react to its `:html` prop changing. Added `:key="sanitizedContent"` so vue-letter remounts on content change.

## Test plan
- [ ] Send a private note → Edit / Delete buttons appear under the bubble
- [ ] Edit → textarea opens, save → bubble updates immediately (no refresh)
- [ ] Delete → confirm dialog → bubble shows the tombstone, Edit/Delete disappear
- [ ] Open the same conversation as a different agent → Edit/Delete are NOT shown on someone else's note
- [ ] `curl -XPUT` a non-private message → 403
- [ ] `curl -XDELETE` another agent's private note (with valid session) → 403
- [ ] After delete, the row still exists in `conversation_messages` with `meta -> 'deleted' = true`


internal/conversation/message.go
```diff
@@ -388,6 +388,40 @@ func (m *Manager) MarkMessageAsPending(uuid string) error {
...
+		return envelope.NewError(envelope.InputError, m.i18n.Ts("globals.messages.notFound", "name", "{globals.terms.message}"), nil)
+	}
+	return nil
+}
+
+// SoftDeletePrivateNote marks a private note as deleted by replacing its
+// content with a tombstone naming the actor and setting `meta.deleted=true`.
+// The row stays in the table so the audit trail is preserved.
+func (m *Manager) SoftDeletePrivateNote(uuid, actorName string) error {
+	tombstone := fmt.Sprintf("This note was deleted by %s", actorName)
```
abhinavxd, line 412: Can we just show "Deleted" (translated) as the bubble content, without the actor name?

The bubble already shows the sender's name above it, and since an agent can only delete their own private notes, the deleter is always the same as the author. 

Showing the name a second time is redundant.

frontend/apps/main/src/features/conversation/message/MessageBubble.vue
```diff
@@ -231,6 +284,83 @@ const retryMessage = (msg) => {
...
+      conversation_uuid: convStore.current.uuid,
+      uuid: props.message.uuid,
+      content: t('conversation.privateNoteDeleted'),
+      meta: { ...(props.message.meta || {}), deleted: true }
+    })
+  } catch (err) {
+    emitter.emit(EMITTER_EVENTS.SHOW_TOAST, {
+      variant: 'destructive',
+      title: t('globals.messages.delete'),
+      description: err?.response?.data?.message || err.message
```
abhinavxd: There's already a utility for this

import { handleHTTPError } from '@shared-ui/utils/http.js'

e.g. usage
```js
    emitter.emit(EMITTER_EVENTS.SHOW_TOAST, {
      variant: 'destructive',
      description: handleHTTPError(err).message
    })
```

cmd/messages.go
```diff
@@ -174,6 +174,82 @@ func handleRetryMessage(r *fastglue.Request) error {
...
+	}
+	if _, err = enforceConversationAccess(app, cuuid, user); err != nil {
+		return sendErrorEnvelope(r, err)
+	}
+
+	msg, err := app.conversation.GetMessage(uuid)
+	if err != nil {
+		return sendErrorEnvelope(r, err)
+	}
+	if msg.ConversationUUID != cuuid || !msg.Private || msg.SenderType != cmodels.SenderTypeAgent || msg.SenderID != user.ID {
```
abhinavxd: The ownership/private-note check is duplicated in `handleUpdatePrivateNote` and `handleDeletePrivateNote`. Worth pulling into a small helper so the rule lives in one place.

abhinavxd:
Added a few comments. Once those are addressed we can get this merged.

abhinavxd:
Thanks, I just tried this PR by running it locally - 

But this doesn't match existing patterns - edit should reuse the rich-text editor (textarea drops the private note formatting), destructive actions use the Dialog component not window.confirm, and message actions belong in a kebab menu or a context menu.

PR #298, opened 2026-04-24
Multi-status conversation list filter

mageaustralia:
## What this changes

The status dropdown on the conversation list becomes multi-select via checkboxes. An agent can tick "Open + Replied" to see all active conversations in one view, rather than flipping the filter back and forth. An "All" option clears every filter.

The single-status case is preserved: pick one status, behaviour is identical to before.

## Wire format

Single status uses the existing `operator: 'equals'` (cheaper plan, unchanged shape). Multi uses `operator: 'in'` with a JSON-stringified array, which `internal/dbutil/builder.go:144` already supports — no backend changes needed.

## A few things flagged in self-review worth calling out

- **`setListStatus` back-compat.** It still accepts a single string and `''` (treated as "All") so existing call sites don't break. Internal state is normalised to an array.

- **`getListStatus` rename.** The old getter doubled as both a value (`if (!getListStatus)`) and a display string. With the value now an array, the truthiness check broke silently. Renamed to `listStatusLabel` (display only) and rewired the two `InboxView.vue` callers to read `conversations.status.length` directly. Decoupling these means the next person to add a derived display value isn't tempted to branch on its content.

- **Last-checked status is disabled** rather than silently ignored on click. Picking zero statuses is what the "All" item is for; making the last one un-clickable avoids the dead-click-feels-broken UX.

- **i18n.** Display strings (`All`, `N statuses`) go through `$t` / the existing `globals.messages.all` key, plus a new `conversation.nStatuses` key with a `{count}` interpolation.

## Test plan
- [ ] Tick a second status (e.g. Replied alongside Open) — list refetches and includes both, label shows "2 statuses"
- [ ] Untick down to one — list refetches with single status; the remaining checkbox becomes disabled
- [ ] Click "All" — every status checkbox unticks, list refetches with no status filter, label shows "All"
- [ ] Refresh — selection state isn't persisted (existing behaviour, unchanged)
- [ ] Switch to a view-based inbox (`/views/:id`) — status dropdown is hidden (existing behaviour)
- [ ] Open InboxView for the first time — defaults to Open as before


abhinavxd:
Not planned closing.

PR #365, opened 2026-06-03
Resolves #363, implement CTRL+M hotkey to access the macro dialog

### LINKED ISSUE #363: [Feature Request] Shortcut to open macros directly
Right now you have to press ctrl + k and then navigate to "apply macros",
Could be a CTRL+ M shortcut for opening macros directly.

vanboom:
CTRL+M for quick access to macro dialog.  It works from either the new conversation or reply conversation paths. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Command dialog keyboard shortcut now reliably clears prior nested command state when opened, preventing stale command behavior.
* **New Features**
  * Macro dialog now preserves and restores the previously selected macro view across open/close cycles for a smoother workflow.
* **Store**
  * Macro view state is now exposed for consistent UI coordination across the app.

### DIFF
--- a/frontend/apps/main/src/features/command/CommandBox.vue
+++ b/frontend/apps/main/src/features/command/CommandBox.vue
@@ -299,6 +299,23 @@ const { Meta_K, Ctrl_K } = useMagicKeys({
 })
 
 watch([Meta_K, Ctrl_K], ([mac, win]) => {
+  setNestedCommand(null)
+  if (mac || win) toggleOpen()
+})
+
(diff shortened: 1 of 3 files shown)


vanboom:
PR #365

abhinavxd:
Found a regression, on the new conversation screen if you hit ctrl+K before ctrl+M it wipes the new-convo flag, so the macro gets treated as a reply and just silently does nothing.

Reproduced it locally.

abhinavxd:
Merging this, seems to work fine now.

Thank you!

PR #366, opened 2026-06-03
Resolves #364, Append macro content into the editors instead of replacing the entire content

### LINKED ISSUE #364: [Feature Request] Allow macros to paste into the current cursor position instead of overwriting the reply message
In Zammad, the user can type :: to select a text snipping for placement into the editor while composing a message.  The macro feature will overwrite anything in the reply box currently, would it feasible to modify it to insert the macro text at the cursor location instead of overwriting the entire reply?

vanboom:
Appending macros allows them to be used as composable text snippets while editing messages.  This works very well in companion with the CTRL+M feature that allows quick access to the macros from the text editor. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed macro content insertion to preserve existing message content instead of replacing it. When applying a macro in conversations or replies, your current text is now appended with the macro content rather than being overwritten.

### DIFF
--- a/frontend/apps/main/src/features/conversation/CreateConversation.vue
+++ b/frontend/apps/main/src/features/conversation/CreateConversation.vue
@@ -503,9 +503,10 @@ const createConversation = form.handleSubmit(async (values) => {
 watch(
   () => conversationStore.getMacro(MACRO_CONTEXT.NEW_CONVERSATION).id,
   () => {
+    const currentValue = form.values?.content || ''
     form.setFieldValue(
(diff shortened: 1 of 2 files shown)


abhinavxd:
We should insert macro content at the cursor position - that's even better.

The current string concat also sometimes leaves a stray <p> on an empty editor inserting through the editor avoids that too.

abhinavxd:
Closing this as thread as unresolved issues, please re-open if needed.

PR #357, opened 2026-06-01
Various IMAP improvements

vanboom:
1. Close #355 by applying filter so deleted emails are not pulled.
1. Close #354 by allowing Fetch to update the seen flag.
1. Close #352 by implementing an Expunge Inbox option on the IMAP channel configuration.

I am not sure I have the capability to update all of the language translations.  How do you typically do that?  AI? 

## Summary by CodeRabbit

* **New Features**
  * Added an option to delete imported emails from the IMAP server after successful import.
  * Added the corresponding inbox setting and description to the manual IMAP configuration form.
  * OAuth-created inboxes keep server-side message deletion disabled by default.

* **Bug Fixes**
  * Improved inbox navigation refresh behavior to update the current conversation list without unnecessary reloads.

PR #397, opened 2026-07-04
feat: show disconnected email inboxes in the UI

### LINKED ISSUE #388: [Feature Request] Show UI messages when a inbox is disconnected and recommend reconnecting the inbox.
Right now if a inbox (OAuth - Google / Mcsft ) disconnects there's no way to tell on the UI if there's an issue without checking logs,

Would be nice to show on the UI and in the inboxes table (Exclamation mark)

Melvin0070:
Closes #388.

OAuth (Google/Microsoft) and IMAP inboxes can silently stop receiving mail when a token is revoked or credentials are rejected — today the only signal is a log line. This surfaces that state so admins know to reconnect.

## Approach

- **Detect** — flag the inbox on IMAP auth / OAuth-refresh failure and clear it on the next successful auth, reusing the existing `TokenRefreshCallback` pattern (email channel → inbox manager). Only credential failures flip it; transient dial/network errors are left alone so the indicator doesn't flap. Status is tracked in memory and seeded from the DB on load, so it writes only on a real transition, not on every poll.
- **Store** — a nullable `inboxes.disconnected_at` column (migration `v2.5.0`). I kept this out of the `config` JSONB since that's encrypted and rewritten on every update, and a column is cleaner to surface/query. Happy to move it into config JSON instead if you'd prefer no schema change.
- **Show** — a warning badge + tooltip in the inboxes table, and the existing OAuth box in the inbox edit form becomes a reconnect banner when disconnected. The reconnect flow itself already existed.

## Testing

`go build ./...`, `go vet`, the email package tests, `eslint`, and `pnpm build:main` all pass. I wasn't able to exercise a real token revocation end-to-end; glad to walk through any manual verification you'd like.

## Notes

- Only `en-US.json` is touched (source locale); other languages come through Crowdin.
- Touches a few files also in #357 (`imap.go`, `models.go`, `EmailInboxForm.vue`), so whichever lands second will need a rebase.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Inbox lists now show connection status for email inboxes.
  * Email inbox settings now display a clear disconnected state with reconnect guidance.

* **Bug Fixes**
  * Inbox connection status is now updated automatically when mail sync disconnects or reconnects.
  * Database upgrades now include the latest schema changes for inbox connection tracking.


Melvin0070:
Thanks for the reviews — both findings addressed in 70fe307:

- **`setConnectionStatus` persistence ordering** — the cached `disconnected` flag is now committed only after the DB write succeeds, so a failed persist is retried on the next poll instead of being swallowed by the transition check.
- **Transient OAuth errors** — a refresh failure now flags a disconnect only when the provider actively rejects our credentials; transport-level failures (network/DNS/timeout, surfaced as `*url.Error`) are left alone, matching the intended "no flap on transient errors" behavior.

abhinavxd:
Hey, thanks for the PR.

I am working whatsapp channel side by side, I ended up fixing this issue there.
Where instead of storing `disconnected_at` in database I've simply save it in a map [here](https://github.com/abhinavxd/libredesk/pull/390/changes#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R121)

PR #438, opened 2026-07-22
seed default business hours and SLA policy on install

### LINKED ISSUE #420: [Feature Request] Fresh install seed a business hours and a SLA policy
One business hours / SLA policy should do.

chahat-101:
Closes #420.

Seeds one default business hours entry (Mon–Fri, 09:00–17:00) and one default
SLA policy (1h first response, 24h resolution).

Follows the existing pattern of a `schema.sql` insert plus a guarded copy in the
current migration. The migration guards on table emptiness rather than name, so
a configured install won't get a stray row and a deleted default won't be
resurrected.

Tested: fresh `--install` + `--upgrade` yields exactly one row in each table.
Rows render and are editable in the admin UI.

## Summary by CodeRabbit

* **New Features**
  * Added default business hours with standard weekday operating times.
  * Added a default SLA policy with first-response and resolution targets.
* **Bug Fixes**
  * Prevented duplicate default records when applying migrations to existing installations.

### DIFF
--- a/internal/migrations/v2.5.0.go
+++ b/internal/migrations/v2.5.0.go
@@ -31,39 +31,5 @@ func V2_5_0(db *sqlx.DB, fs stuffbin.FileSystem, ko *koanf.Koanf) error {
 	if _, err := db.Exec(`CREATE UNIQUE INDEX IF NOT EXISTS index_uniq_conversation_drafts_on_conversation_id_and_user_id_and_type ON conversation_drafts (conversation_id, user_id, type);`); err != nil {
 		return err
 	}
-
-// Seed default business hours and SLA policy. mirrors the INSERTs at the tail of schema.sql.
-// guarded on emptiness, not name, so configured installs aren't given a stray row.
-
-
-	if _, err := db.Exec(`
-		INSERT INTO business_hours ("name", description, is_always_open, hours, holidays)
-		SELECT
-			'Default',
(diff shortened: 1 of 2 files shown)


internal/migrations/v2.5.0.go
```diff
@@ -31,5 +31,39 @@ func V2_5_0(db *sqlx.DB, fs stuffbin.FileSystem, ko *koanf.Koanf) error {
 	if _, err := db.Exec(`CREATE UNIQUE INDEX IF NOT EXISTS index_uniq_conversation_drafts_on_conversation_id_and_user_id_and_type ON conversation_drafts (conversation_id, user_id, type);`); err != nil {
 		return err
 	}
+
+// Seed default business hours and SLA policy. mirrors the INSERTs at the tail of schema.sql.
+// guarded on emptiness, not name, so configured installs aren't given a stray row.
+
+
+	if _, err := db.Exec(`
```
abhinavxd: These migrations will run for anyone upgrading Libredesk, so no migration is needed because they will have already created their own SLA policies and business hours.

We only need the `schema.sql` change since we want to seed the database only during the initial installation.

abhinavxd (changes requested):

internal/migrations/v2.5.0.go
chahat-101: makes sense — dropped the migration, schema.sql only.

Issue #348, opened 2026-05-30
[Feature Request] Google Groups header remapping


mark-ignacio:
We currently use Google Groups to receive customer emails, and we'd like to try out using libredesk by hooking up a GMail account subscribed to that group. We currently auto-label all emails from the group with a filter, so we can have libredesk watch a particular IMAP folder.

libredesk doesn't currently have the ability to do custom handling of headers to remap the to/from fields repsectively. 

From browsing through the relevant code in `imap.go`, I can contribute a PR that implements it! If you're a fan of this idea, would you like me to add this feature?

My idea is to make a new IMAPConfig option called something like `GoogleGroupMode`, which would trickle down to `fetchAndProcessMessages` -> `processEnvelope` and allow it to switch to using the `X-` headers that contain the relevant information.

abhinavxd:
Thanks, happy to take a PR for this!

One suggestion: instead of a Google specific GoogleGroupMode, let's make it a generic optional IMAP config field like original_sender_header (e.g. set to X-Original-Sender).

When set, we use that header's address as the sender, falling back to the normal From when the header is absent. Empty default keeps current behavior for everyone else. That way it also works for almost everything out there and not just google.

And we could also add a hint below "For Google Groups use X-Original-Sender".

For `to` address is this needed? I am not sure.

amaanJvd:
I'd like to take this one, implementing it exactly as you described — a generic optional header rather than anything Google-specific.

**Plan**

- **Config:** add an optional `original_sender_header` field to `IMAPConfig` (`internal/inbox/models/models.go`), empty by default.
- **Ingestion:** in `processEnvelope` (`internal/inbox/channel/email/imap.go`), where `fromAddress` is currently taken from `env.From[0]`, first check the configured header on the incoming message. If `original_sender_header` is set **and** the header is present with a valid address, use that address (and its display name) as the sender; otherwise fall back to the current `From` behavior. Empty config = today's behavior for everyone.
- **Frontend:** add an optional text input in `EditInbox.vue` with a hint below it — *"For Google Groups, set this to X-Original-Sender."*
- **i18n + docs:** add the label/hint strings and a short note in the inbox docs.

**One open question you raised:** you weren't sure about the `to` address. My suggestion is to keep this PR scoped to **sender (`From`) remapping only** — that's what fixes the Google Groups case — and leave `to`/`cc` untouched. If a `to` remap turns out to be needed, it can be a small follow-up with its own header field.

[...]

PR #439, opened 2026-07-23
Optional original-sender header for IMAP inboxes (#348)

amaanJvd:
Implements #348.

Went with the generic approach you suggested rather than anything Google-specific: there's a new optional `original_sender_header` field on the IMAP inbox config. When it's set, we take the address from that header (e.g. `X-Original-Sender` for Google Groups) as the sender, and fall back to the normal `From` whenever the header isn't there. Left empty it changes nothing, so existing inboxes behave exactly as they do today.

I kept this to the `From`/sender side only since that's what actually fixes the Google Groups case. If a `to` remap turns out to be needed we can do it as a small follow-up with its own header field.

A few notes on how it works:

- The configured header name is threaded into the fetch so it's actually requested from the server, then parsed with `net/mail` — handles both a bare `user@example.com` and `Name <user@example.com>`. Anything that doesn't parse just falls back to `From` instead of dropping the message.
- The contact name uses the header's display name when there is one, otherwise the local part of the address.
- Added the field to the inbox form (both auth variants) with a hint pointing Google Groups users at `X-Original-Sender`, plus the en-US strings.

For testing I added a unit test around the header parsing (`extractSenderFromHeader`) covering the display-name, bare-address, missing-header and unparseable cases. `go build ./...` and `go vet ./internal/inbox/...` are clean and the email package tests pass. I didn't spin up the frontend locally, but the new field just mirrors the existing IMAP inputs.

Happy to move the field or adjust anything if you'd rather it sit elsewhere.

## Summary by CodeRabbit

* **New Features**
  * Added an optional IMAP setting to override the effective sender using a configurable original-sender header.
  * Available for both OAuth and manual IMAP inbox setup; when present and valid, emails use the header’s address and display name (otherwise they fall back to the standard sender).
  * Added localized labels, descriptions, and an `X-Original-Sender` example for Google Groups.

* **Bug Fixes**
  * Added validation to reject unsafe/invalid header names.

* **Tests**
  * Added unit coverage for header parsing and header-name validation.

### DIFF
--- a/frontend/apps/main/src/features/admin/inbox/EmailInboxForm.vue
+++ b/frontend/apps/main/src/features/admin/inbox/EmailInboxForm.vue
@@ -295,6 +295,19 @@
           <FormMessage />
         </FormItem>
       </FormField>
+
+      <FormField v-slot="{ componentField }" name="imap.original_sender_header">
+        <FormItem>
+          <FormLabel>{{ $t('admin.inbox.originalSenderHeader') }}</FormLabel>
+          <FormControl>
+            <Input type="text" placeholder="X-Original-Sender" v-bind="componentField" />
+          </FormControl>
+          <FormDescription>
+            {{ $t('admin.inbox.originalSenderHeader.description') }}
+          </FormDescription>
+          <FormMessage />
+        </FormItem>
+      </FormField>
     </div>
 
     <!-- OAuth SMTP Configuration -->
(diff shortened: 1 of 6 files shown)


amaanJvd:
Thanks for the review — pushed a follow-up for two of the three:

- **Sender in message meta:** good catch, that was inconsistent. When the sender is remapped, `meta["from"]` now uses the effective address so it matches the contact (and isn't left empty when the envelope has no `From`).
- **Header name validation:** the configured header is now trimmed and checked (printable ASCII, no spaces/colon) before it goes into `HEADER.FIELDS`; anything invalid is ignored and we fall back to `From` rather than risk a malformed fetch. Added the same trim/validation to the frontend schema too.

On the FETCH `Close()` error propagation — I left that one out on purpose since it's pre-existing in `fetchAndProcessMessages` (the loop already returns `nil` after the stream ends regardless of my change), and validating the header removes the path where this feature could trigger a failed fetch. Happy to fix it here if you'd like, but figured it's a separate robustness change worth its own PR rather than quietly folding it into this one.

`go build ./...`, `go vet`, and the email package tests are still green.

amaanJvd:
Thanks for merging #454 🙏

Whenever you get a chance, this one (the optional original-sender header) is ready too — I pushed fixes for the CodeRabbit feedback a while back (effective sender now consistent in message meta, plus header-name trim/validation), and `go build` / `go vet` / the email package tests are all green. Happy to reshape anything if you'd prefer it done differently.

abhinavxd:
Hey,

I'll check this over weekend.

abhinavxd:
I'll merge this closer to next release, as HelpCenter is releasing next.
This is a risky change and needs to be carefully checked.

PR #450, opened 2026-07-30
Fix live chat image thumbnails for S3 store.

abhinavxd:
## Summary by CodeRabbit

* **New Features**
  * Image attachment previews now prefer per-attachment thumbnail URLs when available.
* **Bug Fixes**
  * If a thumbnail fails to load, the preview falls back to the original image.
  * Stops repeated fallback attempts after an image load error.
* **Refactor**
  * Centralized attachment URL/thumbnail signing and ensured both conversation responses and widget rendering use the same thumbnail/URL logic.


abhinavxd:
@coderabbitai  full review

PR #454, opened 2026-07-31
fix(log): don't drop the error detail in unsnoozeAll log call

amaanJvd:
Small correctness fix in the unsnoozer.

`unsnoozeAll` logged its error with the error passed as a positional argument and no key:

```go
c.lo.Error("error unsnoozing all conversations", err)
```

logf treats variadic fields as key/value pairs and drops the last one when the count is odd (`handleLog`: *"If there are odd number of fields, ignore the last"*). With a single `err` field that's an odd count, so the error was silently discarded — when the unsnooze query fails the log line ended up with no diagnostic detail at all, which makes that background worker hard to debug.

Passed it as an `"error"` key/value, matching the convention used everywhere else in the codebase.

`go build` and `go vet` on the package are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error log formatting when restoring snoozed conversations, making troubleshooting details clearer and more consistent.


abhinavxd:
Thank you

Issue #470, opened 2026-08-05
Mobile support for the agent dashboard: PoC proposal


pat-s:
I run a self-hosted libredesk instance and wanted to be able to triage tickets from my phone.
Rather than open a feature request, I built it and deployed it, so I am able to showcase the result directly instead of discussing about theoretical concepts.

I want to be upfront that I understand the current behaviour is deliberate, not an oversight.
The widget app uses `width=device-width` while the dashboard pins `width=1280, initial-scale=0.29`, and `SmallScreenOverlay.vue` exists with a `scale(3)` tuned specifically to cancel that out.
So the question here is not "you forgot about mobile", it is "would you consider changing your mind, and would you take a PR for it".

Related: #106 asked for this and was closed by the stale bot without a reply, so I could not tell what your position actually is.

## What it looks like

**Screenshots** (based on `v2.6.1`):

<details>
<summary>Details</summary>

<img width="1179" height="1802" alt="Image" src="https://github.com/user-attachments/assets/f84bf3a1-b2a0-4c57-8a78-487f50b43583" />

<img width="1179" height="2166" alt="Image" src="https://github.com/user-attachments/assets/3cd083d8-e8ec-4d43-aabf-5b083d55e3d7" />

<img width="1179" height="2142" alt="Image" src="https://github.com/user-attachments/assets/ec515ef2-6001-4eeb-97f2-0d3c20da9afa" />

</details>

## What works

The full triage loop on a phone, with no pinch-zooming: log in, reach an inbox via the drawer, read the conversation list, open a conversation and read the whole thread, write and send a reply, open an attachment, go back to the list.

## How it is built

The interesting part is how little of this is new code.
Most of the machinery is already in the repo and simply unreachable while the viewport is pinned:

- The mobile drawer already exists. `shared-ui/components/ui/sidebar/Sidebar.vue` renders a `Sheet` when `isMobile`, driven by the `useMediaQuery('(max-width: 768px)')` in `SidebarProvider.vue`, and the `SidebarTrigger`s are already mounted in `PageHeader.vue` and `ConversationList.vue`. It never fires because the media query cannot match at a fixed 1280px layout viewport. The main blocker was `collapsible="none"` on the icon rail in `App.vue`, which short-circuits the Sheet branch.
- The inbox router is already nested. `router/index.js` makes the conversation a real grandchild route with its own URL, so list-to-detail push navigation on mobile needed no router changes at all.
[...]

PR #472, opened 2026-08-06
fix(ui): make dialogs fit a narrow viewport

pat-s:
`DialogContent` and `AlertDialogContent` are `w-full max-w-lg` with no side inset, no `max-h` and no `overflow-y-auto`, and `sm:rounded-lg` only applies from 640px. Below that they render edge-to-edge with square corners, and can't scroll when taller than the viewport.

Adds a 1rem side inset, `max-h-[calc(100dvh-2rem)]` + `overflow-y-auto`, and makes the radius unconditional. At ≥640px `max-w-lg` still governs, so nothing changes there.

Two files, covers all 17 dialog call sites. Refs #470

## Summary by CodeRabbit

* **Style**
  * Improved alert dialogs and standard dialogs with responsive horizontal spacing and sizing.
  * Added viewport-aware maximum heights and vertical scrolling for content-heavy dialogs.
  * Applied rounded corners consistently across screen sizes.

### DIFF
--- a/frontend/shared-ui/components/ui/alert-dialog/AlertDialogContent.vue
+++ b/frontend/shared-ui/components/ui/alert-dialog/AlertDialogContent.vue
@@ -43,7 +43,7 @@ const forwarded = useForwardPropsEmits(delegatedProps, emits)
       v-bind="forwarded"
       :class="
         cn(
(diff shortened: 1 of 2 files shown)


abhinavxd:
Closing as I've moved commits to #473

PR #492, opened 2026-08-14
feat(auth): add SSO-only login mode

### LINKED ISSUE #491: Add an SSO-only authentication mode
## Problem

LibreDesk supports OIDC providers, but local email/password authentication remains available whenever SSO is configured.

Organizations that require authentication exclusively through their identity provider cannot currently hide the password form or prevent direct requests to the local login endpoint.

Hiding the form alone would not be sufficient because local authentication would remain available through the API.

## Proposed behavior

Add an SSO-only authentication mode that can be enabled after at least one active OIDC provider is configured.

When enabled, LibreDesk should:

- Show only enabled OIDC providers on the login page.
- Omit the email/password form, password reset link, and separator.
- Reject local password login requests on the server.
- Keep existing agents and their OIDC email matching behavior unchanged.
- Prevent enabling the mode when no OIDC provider is active.

An installation setting or environment/config option would both work, provided the backend enforces it.

## Recovery

Please retain a documented recovery path for the System administrator, such as a CLI flag or an explicit configuration override.

This avoids permanent lockout when the external identity provider is unavailable or misconfigured.

## Use case

We run LibreDesk behind ZITADEL and want internal and external agents to authenticate exclusively through centralized OIDC.

Password lifecycle, MFA, social identity federation, and account access policies should remain the responsibility of the identity provider.

pat-s:
Closes #491.

## Summary

- Add an `auth.local_login_enabled` configuration option that defaults to enabled for backward compatibility.
- Hide the password form and its separator when local login is disabled.
- Enforce the setting on the login API so bypassing the UI cannot authenticate locally.
- Automatically keep local login available if no OIDC provider is enabled to prevent administrator lockout.
- Document the setting and cover its fallback behavior with tests.

## Verification

- `go test ./...`
- `bun run build:main`

## Summary by CodeRabbit

* **New Features**
  * Added an authentication setting to enable or disable local email/password sign-in.
  * Local sign-in is disabled when configured off and an OIDC provider is enabled.
  * Local sign-in remains available by default or when OIDC authentication is unavailable.
  * The login screen hides local sign-in options when unavailable.
  * Public configuration reports whether local sign-in is enabled.
  * Added guidance for configuring local sign-in and OIDC authentication.

* **Bug Fixes**
  * Blocked local-login requests when local authentication is disabled.
  * Added a localized message directing users to SSO when password login is unavailable.

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -21,6 +21,14 @@ func handleLogin(r *fastglue.Request) error {
 		loginReq loginRequest
 	)
 
+	oidcProviders, err := app.oidc.GetAll()
+	if err != nil {
+		return sendErrorEnvelope(r, err)
+	}
+	if !isLocalLoginEnabled(oidcProviders) {
+		return r.SendErrorEnvelope(fasthttp.StatusForbidden, "Local login is disabled. Sign in with an SSO provider.", nil, envelope.PermissionError)
+	}
+
 	// Decode JSON request.
 	if err := r.Decode(&loginReq, "json"); err != nil {
 		return r.SendErrorEnvelope(fasthttp.StatusBadRequest, app.i18n.T("errors.parsingRequest"), nil, envelope.InputError)
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -21,12 +21,15 @@ func handleLogin(r *fastglue.Request) error {
 		loginReq loginRequest
 	)
 
-	oidcProviders, err := app.oidc.GetAll()
-	if err != nil {
-		return sendErrorEnvelope(r, err)
-	}
-	if !isLocalLoginEnabled(oidcProviders) {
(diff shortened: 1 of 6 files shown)


cmd/login.go
```diff
@@ -21,6 +21,14 @@ func handleLogin(r *fastglue.Request) error {
 		loginReq loginRequest
 	)
 
+	oidcProviders, err := app.oidc.GetAll()
+	if err != nil {
+		return sendErrorEnvelope(r, err)
+	}
+	if !isLocalLoginEnabled(oidcProviders) {
+		return r.SendErrorEnvelope(fasthttp.StatusForbidden, "Local login is disabled. Sign in with an SSO provider.", nil, envelope.PermissionError)
```
abhinavxd: We can translate this in en-us.json

abhinavxd (changes requested):

abhinavxd:
Thank you I'll check this over this week.

abhinavxd:
Thank you

PR #502, opened 2026-08-17
fix(updates): add HTTP timeout and always close body in update checker (#445)

### LINKED ISSUE #445: [Bug] Update checker has no HTTP timeout and leaks response bodies on error paths
**Version:**
- libredesk: built from main, commit 1a0a78f4 (2026-07-25)

`cmd/updates.go` checks for new releases on a 24h ticker. Two problems:

1. No HTTP timeout (main issue). `fnCheck` uses `http.Get`, which uses
`http.DefaultClient` with a zero timeout:

    resp, err := http.Get(updateCheckURL)

The check runs synchronously in the ticker loop:

    for range ticker.C {
        fnCheck()
    }

If the endpoint accepts the connection but never responds (hung LB,
blackholing proxy, captive portal), `http.Get` blocks forever. The loop
never advances, so update checking is dead until restart. The call never
returns, so the error branch never runs and nothing is logged -- no way to
tell from the outside. The admin just never sees the update banner and
assumes they're current.

2. Response body not closed on error paths (secondary):

    if resp.StatusCode != 200 {
        ...
        return // body not closed
    }
    b, err := io.ReadAll(resp.Body)
    if err != nil {
        ...
        return // body not closed
    }
    resp.Body.Close() // only reached on success

The non-200 and read-error returns skip Close(), leaking the connection.

Repro for the hang: point the update URL at a host that accepts the TCP
connection but never responds (e.g. a firewalled port). The checker blocks
indefinitely, logs nothing, and never recovers.

**Suggested fix:**

Use a client with a timeout (already done in `internal/ai/openai.go`):

    client := &http.Client{Timeout: 10 * time.Second}

and move `resp.Body.Close()` to a defer after the error check.

Note: `cmd/updates.go` is adapted from listmonk under AGPL. The same two
issues exist in listmonk's current `cmd/updates.go`, so there's no upstream
fix to port.

I'd be happy to take this if you can assign it to me.

**Logs / Screenshots:**
N/A -- the symptom is the absence of any log output when the endpoint hangs.

chiliec:
## What

Closes #445.

`cmd/updates.go` checks for new releases on a 24h ticker. Two problems:

1. **No HTTP timeout (main issue).** `fnCheck` used `http.Get`, which uses `http.DefaultClient` with a zero timeout, and it runs synchronously in the ticker loop:

   ```go
   for range ticker.C {
       fnCheck()
   }
   ```

   If the endpoint accepts the connection but never responds (hung LB, blackholing proxy, captive portal), `http.Get` blocks forever. The loop never advances, so update checking is dead until restart — and since the call never returns, the error branch never runs and nothing is logged.

2. **Response body not closed on error paths (secondary).** `resp.Body.Close()` was only reached on the success path, so the non-200 and read-error returns leaked the connection.

## Fix

Extract the fetch/parse into a package-level `fetchAppUpdate(client *http.Client, url string)`:

- the caller passes `&http.Client{Timeout: updateCheckTimeout}` (10s), so a single check can no longer block forever;
- the body is closed via `defer resp.Body.Close()`, covering every return path;
- taking the client and URL as parameters makes it testable against a stub server.

`fnCheck` now just calls it and applies the version comparison as before — no behavioural change on the happy path.

## Tests

Added `cmd/updates_test.go`:
- **success** — parses the manifest and returns the release version;
- **non-200 closes body** — wraps the transport to observe `Close()`, asserting the body is closed on the error path (this is the leak regression);
- **timeout** — a server that never responds; the call returns within the client timeout instead of blocking.

**Verified RED→GREEN:** reverting to the old body-handling (close only on success) makes the non-200 test fail with `response body was not closed on the non-200 path (leak)`; the fix makes it pass.

## Validation (real results, Go 1.24)

- `go test ./cmd/ -run TestFetchAppUpdate`: **3/3 pass**
- `go build ./cmd/`: **success**
- `gofmt -l` and `go vet ./cmd/`: clean

First-time contributor here — happy to adjust the timeout value or the helper's shape if you'd prefer.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved update checks with a 10-second timeout.
  - Update responses now handle unsuccessful server responses more reliably.
  - Update information is processed more efficiently and consistently.

- **Tests**
  - Added coverage for successful update retrieval.
  - Added coverage for server errors and invalid response statuses.

### DIFF
--- a/cmd/updates.go
+++ b/cmd/updates.go
@@ -7,7 +7,6 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"regexp"
 	"time"
@@ -17,8 +16,6 @@ import (
 
 const updateCheckURL = "https://updates.libredesk.io/updates.json"
 
-// updateCheckTimeout bounds a single update check so a hung or blackholing
-// endpoint cannot block the checker loop forever (see #445).
 const updateCheckTimeout = 10 * time.Second
 
 type AppUpdate struct {
@@ -49,8 +46,10 @@ func checkUpdates(curVersion string, interval time.Duration, app *App) {
 	// Strip -* suffix.
 	curVersion = reSemver.ReplaceAllString(curVersion, "")
 
+	client := &http.Client{Timeout: updateCheckTimeout}
+
 	fnCheck := func() {
-		out, err := fetchAppUpdate(&http.Client{Timeout: updateCheckTimeout}, updateCheckURL)
+		out, err := fetchAppUpdate(client, updateCheckURL)
 		if err != nil {
 			app.lo.Error("error checking for app updates", "err", err)
 			return
@@ -84,10 +83,6 @@ func checkUpdates(curVersion string, interval time.Duration, app *App) {
 	}
 }
 
-// fetchAppUpdate fetches and parses the update manifest using the given client.
-// The response body is always closed, including on the non-200 and read-error
-// paths that previously leaked it (see #445). The caller supplies the client so
-// a timeout can be enforced (and so this can be tested against a stub server).
 func fetchAppUpdate(client *http.Client, url string) (*AppUpdate, error) {
 	resp, err := client.Get(url)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


abhinavxd:
@coderabbitai full review

abhinavxd:
Thank you

PR #539, opened 2026-08-28
Fixes thumbnail rotation issues

### LINKED ISSUE #538: [Possible Bug] Thumbnail orientation
**Version:**
 - libredesk: 2.8.0
 
When I add a portrait photo to the live chat conversation, it sends well but the thumbnail is bad oriented. When I click on it, it opens with the well orientation.
I use S3 method as file storage

**Logs / Screenshots:**

Image
Image

eozsahin1993:
Fixes thumbnail rotation issues

Adding `imaging.AutoOrientation(true)` ensures that the image is decoded in the correct orientation based on the EXIF data without making assumptions.

Manually verified the fix after the change and confirmed that it's working as expected

Fixes #538

Screenshot 2026-08-28 at 12 19 54 PM

## Summary by CodeRabbit

* **Bug Fixes**
  * Thumbnails now automatically respect an image’s EXIF orientation.
  * Thumbnails are generated in PNG format for more consistent output.


abhinavxd:
Thank you

PR #542, opened 2026-08-28
Add a per-user default inbox view

### LINKED ISSUE #541: Per-user default inbox view
## Why

Agents who work a shared pile often want LibreDesk to open on **All** conversations rather than **My Inbox**. Assigned-first is the right default for desks that assign work, but it should not be the only option.

## Proposal

- Each agent can choose a **default inbox**: My Inbox (`assigned`), Mentions (`mentioned`), Unassigned (`unassigned`), or All (`all`).
- The default remains **My Inbox**, so existing behaviour is unchanged.
- Visiting `/inboxes` (and logging in without a `next` URL) lands on that inbox.
- The selected inbox is **pinned to the top of the inbox nav** so the list matches where you arrived. The other three keep their current relative order. Team inboxes stay in their own block.

This is a **per-user** preference on the profile page, not a global admin setting.

## Out of scope

- Reordering team inboxes or custom views
- Changing what the four inboxes mean
- Remembering the last visited conversation (the existing `lastInboxPath` rail behaviour can stay as it is)

Sending a PR with my proposed solution.

josephsellers:
Closes https://github.com/abhinavxd/libredesk/issues/541

Each agent can choose which inbox opens when they sign in: My Inbox, Mentions, Unassigned, or All. The default stays My Inbox, so existing behaviour is unchanged.

The selected inbox is pinned to the top of the inbox nav so the list matches where you arrived. The other three keep their current relative order. Team inboxes are unchanged.

- Profile page: **Default inbox** select (saves immediately)
- `PUT /api/v1/agents/me/default-inbox`
- Column `users.default_inbox` (migration `v2.9.0`), default `assigned`
- Login without a `next` URL, and a bare `/inboxes` / `/inboxes/assigned` load after the user is fetched, land on that inbox
- Deep links to a conversation are not rewritten

Happy to drop the nav pin and ship landing-only if you would rather that as a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added a profile setting to choose a default inbox: Assigned, Mentioned, Unassigned, or All.
  * Users now land in their selected inbox after login and when opening the inbox.
  * Inbox navigation prioritizes the selected default inbox.
  * Added validation and success/error notifications when saving the preference.
* **Bug Fixes**
  * Invalid or inaccessible inbox selections are rejected and safely fall back to Assigned.
  * Default inbox preferences are preserved across sessions.


abhinavxd:
Thanks for the PR,

Going to pass on this one though. The gain is one click after login, and the cost is a schema column, a new endpoint, sidebar reordering, and changes to the login redirect path - more surface than the preference justifies.

josephsellers:
Fair enough, thanks for reviewing. 

Please bear in mind though that for users who never assign tickets but deal with them as a unified inbox (me!) they are greeted with a blank page and have to navigate elsewhere every time.

Perhaps at some point there will be a way to better integrate some alternate preferences for this.

Issue #506, opened 2026-08-18
[Possible Bug] Is there a limit to size of reply?  Editor stops accepting ENTER carriage return.


vanboom:
**Version:**
 
**Description of the bug and steps to reproduce:**

This feels random, but after some typing especially when building a numbered list, the editor stops accepting the ENTER key to create a new line.  A box showing "No results found" appears in the lower left corner of the reply editor.    We can continue to type characters, but it will not accept the ENTER key.  

Curious if anyone can reproduce this.   We started in Firefox, but even when we reload or open the conversation in Chrome, the editor is still stuck in this condition.  

Is there a character limit?  Curious!

eozsahin1993:
You must be creating a mention by typing `@` which won't let you accept the ENTER when there are no results. The reason that must be the case is because `No results found` will only be displayed with mentions in the reply editor. If this is the issue you are seeing, I can create a fix for it.

PR #543, opened 2026-08-28
Fix: disable mentions for replies in tiptap

eozsahin1993:
Fix: disable mentions for replies in tiptap

Found an issue with mentions in the reply editor where suggestions always returned empty results by design. 
```
// ReplyBoxContent.vue

const getSuggestions = async (query) => {
  // Only show suggestions in private note mode
  if (messageType.value !== 'private_note') {
    return []
  }
  ...
```
However, the `mentionSuggestions` stills runs and shows empty results. With this fix, we no longer show the pop up at all.

Manually tested the behavior locally and no-longer seeing the mention pop-up.

Came across this issue when debugging a semi related issue #506. 

## Summary by CodeRabbit

- **New Features**
  - Added configurable mention support for the conversation editor.
  - Mentions can now be enabled or disabled based on the editor’s configuration.
  - Mention suggestions are suppressed when mentions are disabled, preventing unintended mention interactions.
- **Bug Fixes**
  - Improved consistency between the conversation editor’s mention settings and the behavior shown while composing messages.

### DIFF
--- a/frontend/apps/main/src/components/editor/ConversationEditor.vue
+++ b/frontend/apps/main/src/components/editor/ConversationEditor.vue
@@ -84,6 +84,7 @@ const { editor, extractMentions, focus } = useTextEditor({
   isInlineEnabled: () => props.enableInlineImages,
   linkedModel: props.linkedModel,
   getSuggestions: props.getSuggestions,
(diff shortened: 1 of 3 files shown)


abhinavxd:
Thank you

PR #544, opened 2026-08-29
api: accept source_id on conversation and message endpoints

### LINKED ISSUE #545: API: accept source_id on conversation and message endpoints
## Problem

Libredesk builds every outgoing reply's `References` and `In-Reply-To` headers from the `source_id` of prior messages (`BuildEmailThreadingHeaders`). `source_id` is only ever set on the IMAP ingestion path:

```go
// internal/inbox/channel/email/imap.go
SourceID: null.StringFrom(messageID),
```

Neither `POST /api/v1/conversations` nor `POST /api/v1/conversations/{cuuid}/messages` accepts it (see `createConversationRequest` and `messageReq`). A desk fed over the API therefore references only its own previous sends, and the customer's mail client has nothing of theirs to thread on. Every reply arrives looking standalone.

Today the only way to set `source_id` from outside the IMAP path is a direct write to Libredesk's database. That is fragile (an upstream migration renaming or retyping `source_id` breaks it silently), it forces API consumers to hold database credentials they should not need, and it is invisible to upstream.

## Proposal

Add an optional `source_id` to both request structs and store it on the created contact message. The change is small and mirrors what the IMAP path already does.

- `createConversationRequest` and `messageReq` gain an optional `source_id` string.
- It is normalized (whitespace and angle brackets stripped) and stored on the created contact message, since `source_id` is stored unbracketed and the brackets are re-added by `BuildEmailThreadingHeaders`.
- An empty value leaves the column `NULL` as before, so existing callers are unaffected.
- It is honored only for contact-originated messages (the `initiator: "contact"` case on create, and `sender_type: "contact"` on append). Agent replies generate their own `source_id` in `QueueReply` as today.

## Why it fits libredesk

This is needed by any inbox fed over the API rather than IMAP. The `source_id` column and the threading logic already exist; this only opens the existing mechanism to API consumers. It is purely additive: a new optional field, no schema change, no behavior change for callers that do not send it.

## Implementation

- A shared `stringutil.NormalizeMessageID` helper keeps the normalization in step with what the IMAP path already does (the mail library strips brackets there; the HTTP API has to do it itself).
- `CreateContactMessage` takes the `sourceID` argument and sets `SourceID: null.NewString(sourceID, sourceID != "")`, matching the codebase's existing optional-field idiom.
- The two handler call sites pass `req.SourceID` through.

I have a draft open at #544.

malpou:
Closes #545.

## What

Add an optional `source_id` to `POST /api/v1/conversations` and `POST /api/v1/conversations/{cuuid}/messages`, stored on the created contact message. This mirrors the IMAP ingestion path, which is the only place `source_id` is otherwise populated.

## Why

Libredesk builds every outgoing reply's `References` and `In-Reply-To` headers from the `source_id` of prior messages (`BuildEmailThreadingHeaders`). `source_id` is only ever set on the IMAP ingestion path:

```go
// internal/inbox/channel/email/imap.go
SourceID: null.StringFrom(messageID),
```

Neither API endpoint accepts it. A desk fed over the API therefore references only its own previous sends, and the customer's mail client has nothing of theirs to thread on. Every reply arrives looking standalone.

This is an obviously correct addition that any API-fed inbox needs. Today the only way to set `source_id` from outside the IMAP path is a direct database write, which is fragile and invisible to upstream.

## What it does

- `createConversationRequest` and `messageReq` gain an optional `source_id`.
- The value is normalized (whitespace and angle brackets stripped) and stored on the created contact message, since `source_id` is stored unbracketed and the brackets are re-added by `BuildEmailThreadingHeaders`.
- Values containing line breaks are rejected (normalized to empty) so they cannot produce malformed `In-Reply-To`/`References` headers on outbound replies; an empty result leaves the column `NULL`.
- An empty value leaves the column `NULL` as before, so existing callers are unaffected.
- It is honored only for contact-originated messages (the `initiator: "contact"` case on create, and `sender_type: "contact"` on append). Agent replies generate their own `source_id` in `QueueReply` as today.

The normalization is a shared `stringutil.NormalizeMessageID` helper so it stays in step with what the IMAP path already does.

## Testing

- `go build ./...`
- `go test ./internal/stringutil/ ./internal/conversation/ ./cmd/`
- New `TestNormalizeMessageID` pins the bracket/whitespace stripping and the line-break rejection.

## Compatibility

Purely additive: a new optional field. No schema change (the column already exists). No behavior change for callers that do not send `source_id`.

### DIFF
--- a/cmd/conversation.go
+++ b/cmd/conversation.go
@@ -58,6 +58,7 @@ type createConversationRequest struct {
 	Content          string         `json:"content"`
 	Attachments      []int          `json:"attachments"`
 	Initiator        string         `json:"initiator"` // "contact" | "agent"
+	SourceID         string         `json:"source_id"` // RFC 5322 Message-ID of the inbound message; stored on the created contact message so replies thread on it. Contact-initiated only.
 	CustomAttributes map[string]any `json:"custom_attributes"`
 }
 
@@ -882,7 +883,7 @@ func handleCreateConversation(r *fastglue.Request) error {
 		}
 	case umodels.UserTypeContact:
 		// Create contact message.
-		if _, err := app.conversation.CreateContactMessage(media, contact.ID, conversationUUID, req.Content, cmodels.ContentTypeHTML, true); err != nil {
+		if _, err := app.conversation.CreateContactMessage(media, contact.ID, conversationUUID, req.Content, cmodels.ContentTypeHTML, true, req.SourceID); err != nil {
 			// Delete the conversation if message creation fails.
 			if err := app.conversation.DeleteConversation(conversationUUID); err != nil {
 				app.lo.Error("error deleting conversation", "error", err)
--- a/cmd/messages.go
+++ b/cmd/messages.go
@@ -22,6 +22,7 @@ type messageReq struct {
 	SenderType  string                 `json:"sender_type"`
 	Mentions    []cmodels.MentionInput `json:"mentions"`
 	EchoID      string                 `json:"echo_id"`
+	SourceID    string                 `json:"source_id"` // RFC 5322 Message-ID of the inbound message; stored on the created contact message so replies thread on it. Contact sender only.
 }
 
 // handleGetMessages returns messages for a conversation.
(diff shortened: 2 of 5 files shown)


malpou:
@coderabbitai review

malpou:
@coderabbitai full review

abhinavxd:
Thank you

PR #547, opened 2026-08-29
settings: reload auth on general settings update (narrow fix)

### LINKED ISSUE #546: Settings changes do not propagate to subsystems that snapshot them (OIDC redirect_uri is the case that bit us)
## Problem

Changing a setting in **Admin → Workspace → General** does not reliably reach subsystems that captured that setting at construction time. Libredesk has three patterns for consuming settings, two safe and one not, and the unsafe one is used in at least two places.

The instance that cost us most of an evening: changing **Root URL** leaves the OIDC provider's redirect URI stale until the pod restarts, so SSO keeps sending the old URL and the identity provider rejects it with

```
{"error":"invalid_request","error_description":"The requested redirect_uri is missing in the client configuration."}
```

The database held the right `app.root_url`, the IdP client had the right `redirect_uris`, both agreed, and login still failed, because the running process was building the URL from a value cached at boot.

## The general problem

Verified in v2.8.0. Three ways a setting reaches a consumer:

| Pattern | Example | Safe? |
|---|---|---|
| Read `app.consts.Load()` per request | `cmd/oauth.go:65`, `cmd/helpcenter.go:1192`, `cmd/media.go:98` | **Yes** -- `reloadSettings` re-runs `initConstants` and stores a new value |
| Pass a **closure** that fetches on call | media store, `cmd/init.go:579` -- `rootURL := func() string { ... settings.GetAppRootURL() }` | **Yes** |
| **Snapshot the value into a long-lived struct** | `buildProviders`, `cmd/init.go:888` -- `RedirectURL: config.RedirectURI` | **No** -- stale until something explicitly rebuilds it |

The third pattern is the bug class. It is not inherently wrong -- it just creates a dependency on an explicit reload, and the settings handler does not know about it.

## Known instances

1. **OIDC redirect URI (the one that bit us).** `buildProviders` (`cmd/init.go:872`) computes `RedirectURL` once from `rootURL + "/api/v1/oidc/%d/finish"` (the URI is assembled in `internal/oidc/oidc.go:85`). `reloadAuth` (`cmd/init.go:858`) rebuilds the providers and *is* called correctly on OIDC create/update (`cmd/oidc.go:62`, `:99`) -- but the settings handler (`cmd/settings.go:69-89`) calls `reloadSettings`, then re-inits i18n inline if `app.lang` changed, then `reloadTemplates`, and **never `reloadAuth`**. Root URL is therefore the one input to the redirect URI with no reload path at all.

2. **Template funcMap.** `getTmplFuncs` (`cmd/init.go:456`) closes over a captured `consts` value rather than calling `app.consts.Load()`, so `RootURL`, `SiteName`, `LogoURL` and `FaviconURL` in templates are all snapshots. This one happens to be saved because `reloadTemplates` rebuilds the map -- but only by coincidence of ordering, not by design.

## Suggested fix, in order of preference

**1. Make the settings handler reload everything that depends on settings.** The narrow fix -- adding `reloadAuth(app)` to `cmd/settings.go` -- closes the reported bug but leaves the class open: the next subsystem to snapshot a setting reintroduces it, silently.

Better: a single `reloadAll(app)` (or an explicit, named list) invoked by the settings handler, so "what must be rebuilt when settings change" is written down in one place rather than implied by which handler happens to call what.

Note the handler already has the right instinct for i18n -- it compares `oldLang` to the new value and only re-inits when it changed. Extending that shape (reload X when the settings X depends on actually changed) to auth would be consistent with what is there, and cheaper than reloading everything on every save.

**2. Prefer live lookup over snapshots.** Where a value is read rarely (an OIDC redirect is computed per login, not per request), reading it live removes the reload requirement entirely. `buildProviders` could hold the root URL as a closure, matching what the media store already does two hundred lines earlier in the same file. That is the pattern that cannot go stale.

**3. Guard the remaining snapshots.** If a value must be cached, the reload that refreshes it should be adjacent to the code that caches it -- or asserted in a test -- so a future setting cannot quietly acquire the same problem.

## Why this is worth fixing upstream

The failure is silent and misattributed. Nothing logs a warning, nothing looks stale, and the error surfaces at the *identity provider* -- so the natural instinct is to debug the IdP's client configuration, which is correct and will keep looking correct however long you stare at it. There is also no UI hint that Root URL needs a restart to take effect.

It affects anyone running Libredesk behind a hostname configured after first boot, i.e. essentially every containerised deployment.

## Workaround until then

Restart after changing any setting a subsystem may have snapshotted:

```bash
# restart the libredesk process / pod
```

malpou:
Closes #546 (narrow option).

This is the narrow fix for #546. A second PR with the preferred option (live lookup instead of a snapshot) is at #548, so the maintainer can choose the approach. See the comment on this PR.

## What

Add `reloadAuth(app)` to `handleUpdateGeneralSettings`, after the existing `reloadSettings`, i18n re-init and `reloadTemplates`. Also make `reloadAuth` return the `buildProviders` error instead of calling `log.Fatalf`, so a settings save whose only fault is a provider that cannot be built at runtime returns a 500 envelope rather than crashing the process.

## Why

The general settings handler reloads settings, i18n and templates, but not auth. OIDC providers snapshot `app.root_url` into their `RedirectURL` at construction time (`buildProviders`, `cmd/init.go:872`), so changing Root URL in Admin -> Workspace -> General leaves the redirect URI stale until the process restarts. SSO then keeps sending the old `redirect_uri` and the identity provider rejects it:

```
{"error":"invalid_request","error_description":"The requested redirect_uri is missing in the client configuration."}
```

`reloadAuth` already exists and is called correctly on OIDC create/update (`cmd/oidc.go:62`, `:99`); it just has no path from the settings handler.

## Scope

This closes the reported bug (OIDC redirect URI going stale on Root URL change) but does not address the wider class: any other subsystem that snapshots a setting reintroduces it silently. #546 proposes two alternatives -- a single `reloadAll(app)` so the dependency list is written in one place, and live lookup over snapshots, which removes the reload requirement entirely. This PR is the narrow one-liner (plus the fatal-to-error fix the reload path needed); the preferred option is in #548.

## Testing

- `go build ./...`
- `go test ./cmd/`

There is no existing test harness for the settings handler (it requires a full `*App` and a database), so this change is verified structurally: the handler now calls `reloadAuth`, which rebuilds providers from the current `app.root_url`. The `reloadAuth` fatal-to-error change is covered by its three existing callers already handling the returned error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication providers now immediately use updated Root URL settings, including correctly rebuilt OIDC redirect URIs.
  * Improved error handling when authentication settings cannot be reloaded, with failures reported without terminating the application.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -860,7 +860,8 @@ func reloadAuth(app *App) error {
 	app.lo.Info("reloading auth manager")
 	providers, err := buildProviders(app.oidc)
 	if err != nil {
-		log.Fatalf("error reloading auth: %v", err)
+		app.lo.Error("error reloading auth", "error", err)
+		return err
 	}
 	if err := app.auth.Reload(auth_.Config{Providers: providers}); err != nil {
 		app.lo.Error("error reloading auth", "error", err)
--- a/cmd/settings.go
+++ b/cmd/settings.go
@@ -86,6 +86,14 @@ func handleUpdateGeneralSettings(r *fastglue.Request) error {
 		app.lo.Error("error reloading templates", "error", err)
 		return sendErrorEnvelope(r, envelope.NewError(envelope.GeneralError, app.i18n.T("globals.messages.somethingWentWrong"), nil))
 	}
+
+	// Rebuild auth providers, which snapshot settings such as app.root_url into
+	// their RedirectURL at construction time. Without this, a Root URL change
+	// leaves OIDC redirect URIs stale until the process restarts.
+	if err := reloadAuth(app); err != nil {
+		app.lo.Error("error reloading auth", "error", err)
+		return sendErrorEnvelope(r, envelope.NewError(envelope.GeneralError, app.i18n.T("globals.messages.somethingWentWrong"), nil))
+	}
 	return r.SendEnvelope(true)
 }
 


malpou:
Two PRs are open for #546 so the maintainer can pick the approach; this is one of them.

- #547 -- the narrow one-liner: add `reloadAuth(app)` to the general settings handler. Closes the reported bug (OIDC redirect URI no longer goes stale on Root URL change) but leaves the bug class open: the next subsystem to snapshot a setting reintroduces it silently.
- #548 -- the preferred option: resolve the OIDC redirect URL live (closure over the current `app.root_url`) so it cannot go stale at all, matching the media store's existing `rootURL` closure. Removes the reload requirement for Root URL entirely; the settings handler no longer needs to know that auth depends on it.

Both build and pass `go test ./...`. Neither is intended to merge in isolation of the other -- please pick one and I'll close the other.

malpou:
@coderabbitai review

malpou:
@coderabbitai full review

malpou:
@coderabbitai full review

malpou:
@coderabbitai full review

abhinavxd:
Closing in favour of #548

PR #549, opened 2026-08-30
feat: add separate permission for private conversation notes

LaQuay:
Hello, this PR adds a dedicated `messages:write_private` permission for creating private conversation notes.

Previously, `messages:write` controlled both customer-facing replies and private notes, making it impossible to create collaboration-only roles. Roles can now be configured independently to:

- send public replies
- create private notes
- do both
- do neither

The permission is also applied to private-note macro and automation actions and AI-generated conversation summaries.

## Backend

The existing conversation message endpoint now selects the required permission based on message visibility:

- public reply: `messages:write`
- private note: `messages:write_private`

No new endpoint or breaking API change is introduced.

## Frontend

The conversation composer now reflects both permissions independently:

- users with both permissions see both composer modes
- users with only `messages:write` can only send replies
- users with only `messages:write_private` can only create private notes
- users with neither permission do not see the composer

Stored draft types are validated against the user's current permissions, so they cannot restore an unauthorized composer mode.

Copilot reply and private-note actions are also displayed and guarded independently.

The new permission is available in the role editor with the English label:

> Send private notes in conversations

## Migration

Adds the idempotent `v2.9.0` migration.

For backwards compatibility, the migration grants `messages:write_private` to every existing role that already has `messages:write`.

Roles without `messages:write` are left unchanged, and running the migration repeatedly does not create duplicate permissions.

The default Agent and Admin roles include the new permission on fresh installations.

## Alternate paths reviewed

- Direct API message creation uses the same enforced message endpoint.
- Macros distinguish between public replies and private-note actions.
- Automation action permission metadata uses the new permission for private notes.
- AI conversation summaries require the private-note permission.
- Copilot-created private notes pass through the enforced message endpoint.
- Draft endpoints only persist composer state and do not create messages.
- No separate websocket or bulk message-creation path was found that bypasses the existing endpoint.
- Internal system-generated message behaviour was intentionally left unchanged.

## Tests

Added coverage for all four permission combinations:

1. both permissions
2. public-message permission only
3. private-note permission only
4. neither permission

Also added migration coverage verifying that:

- roles with the old permission receive the new permission
- roles without the old permission remain unchanged
- repeated migration execution does not add duplicates

Frontend Cypress coverage verifies the API authorization result and composer behaviour for each permission combination.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated permission for sending private notes.
  * Added permission-aware controls for replies, private notes, summaries, Copilot actions, and macros.
  * Added private-note permission configuration to role management.
  * Existing eligible roles automatically receive private-note access.

* **Bug Fixes**
  * Prevented unauthorized message and private-note sending through the interface and API.

* **Tests**
  * Added coverage for permissions, roles, migrations, macros, and end-to-end access.

### DIFF
--- a/cmd/handlers.go
+++ b/cmd/handlers.go
@@ -72,7 +72,7 @@ func initHandlers(g *fastglue.Fastglue, hub *ws.Hub) {
 	g.GET("/api/v1/conversations/{uuid}/transcript", perm(handleDownloadConversationTranscript, "messages:read"))
 	g.POST("/api/v1/conversations/{cuuid}/messages", auth(handleSendMessage))
 	g.PUT("/api/v1/conversations/{cuuid}/messages/{uuid}/retry", perm(handleRetryMessage, "messages:write"))
-	g.DELETE("/api/v1/conversations/{cuuid}/messages/{uuid}", perm(handleDeleteMessage, "messages:write"))
+	g.DELETE("/api/v1/conversations/{cuuid}/messages/{uuid}", perm(handleDeleteMessage, "messages:write_private"))
 	g.POST("/api/v1/conversations", perm(handleCreateConversation, "conversations:write"))
 	g.PUT("/api/v1/conversations/{uuid}/custom-attributes", auth(handleUpdateConversationCustomAttributes))
 	g.PUT("/api/v1/conversations/{uuid}/contacts/custom-attributes", auth(handleUpdateContactCustomAttributes))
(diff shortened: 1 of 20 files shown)


abhinavxd:
@coderabbitai The macro.js 46-51 finding is already the current behavior: `macroOptions` maps each macro to a clone (`{...macro, actions: macro.actions.filter(...)}`) and never mutates `macroList`, so original actions arrays stay intact. No change needed.

abhinavxd:
Thank you

PR #567, opened 2026-09-05
livechat: add pattern validation for pre-chat form fields

### LINKED ISSUE #564: Pre-chat form: inline intake at handoff, and pattern validation on fields
### What

Two related additions to the livechat pre-chat form.

**1. An inline intake form at handoff time.** Instead of collecting contact details before the conversation starts, collect them at the moment the assistant hands off to a human, rendered as a bubble inside the chat. Configured as `prechat_form.on_handoff` on the livechat inbox.

**2. Pattern validation on pre-chat fields.** `pattern` (a regex) and `pattern_message` (the error shown when it fails) on `PreChatFormField`, both optional.

### Why the form moves to handoff

A pre-chat form asked up front is a wall in front of a question the assistant can usually answer on its own. Every visitor pays the cost; only the ones who end up needing a human get anything back for it. The ones who abandon at the form never got an answer at all.

Asking at handoff inverts that. The visitor has already had the conversation, already knows the assistant could not help, and is being asked for an email so a person can reply. The ask is motivated at the point it is made.

The stronger the knowledge base, the worse the up-front form looks, because the share of visitors who never needed a human keeps rising.

### Why pattern validation

Plenty of deployments key a conversation to an identifier the visitor types in: an account number, an order reference, a postcode. Today there is no way to say what shape it should be, so a typo becomes a ticket nobody can match to an account, and it costs an agent a round trip to sort out.

`^\d{8}$` plus a message saying what the field wants turns that into a corrected keystroke. Both fields are optional and the default is unchanged.

### How

`internal/inbox/channel/livechat`: `Pattern` and `PatternMessage` on `PreChatFormField`; `on_handoff` on the pre-chat form config.

`frontend/apps/widget`: an `IntakeFormBubble.vue` rendered in the message stream, and the pattern check in `preChatFormSchema.js`.

Running in production on a fork since August.

### Before I open a PR

These are two separate concerns and I would send them as two PRs, smallest first. The pattern validation is small and self-contained. The handoff form is the larger of the two and changes a default behaviour, so it is the one worth telling me to drop if it is not where you want the widget to go.

pujitha24:
Motivation:
Pre-chat form fields today only validate by type (email format, URL
format, etc). Deployments that key a conversation to a free-text
identifier such as an account number or order reference have no way
to shape-check it, so a typo becomes a submission nobody can match to
an account.

Approach:
Add optional `Pattern` (a regex) and `PatternMessage` (its error text)
to `PreChatFormField` in internal/inbox/channel/livechat/livechat.go.
Both are empty by default, so existing configs are unaffected. In the
widget's preChatFormSchema.js, when a field carries a pattern, chain a
`.regex()` check onto its zod schema using the same fall-through path
already used for email/date/text/list fields; an invalid regex string
is caught and ignored rather than breaking form validation for every
visitor. Optional empty fields keep bypassing the pattern check, same
as they already bypass other format checks. Matches the existing
codebase convention that field-shape validation is enforced
client-side in the widget, not duplicated server-side.

There is no admin UI to set these two fields yet; for this first PR
they're set directly on the inbox's stored config (e.g. via the API),
matching how the issue describes this half of the change.

Validation:
- go build ./... and go vet ./... pass.
- go test -count=1 ./... passes, aside from a pre-existing failure in
  internal/stringutil (TestNormalizeTimezoneFallsBackToUTC) that
  reproduces identically on the unmodified tree (verified via git
  stash) and is unrelated to this change.
- cd frontend && pnpm exec vitest run: 778 tests pass, including 3 new
  cases added to preChatFormSchema.test.js covering a rejected
  pattern, an optional empty value skipping the pattern, and an
  invalid pattern string being ignored instead of throwing.
- cd frontend && pnpm exec eslint on the two changed frontend files:
  clean.

Report: https://github.com/abhinavxd/libredesk/issues/564
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>
Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #564

## Summary by CodeRabbit

* **New Features**
  * Added configurable pattern validation for pre-chat fields.
  * Custom validation messages can now be displayed when values do not match the configured pattern.
  * Optional fields skip pattern validation when left empty.

* **Bug Fixes**
  * Invalid pattern configurations are safely ignored instead of causing errors.

### DIFF
--- a/frontend/apps/widget/src/components/preChatFormSchema.js
+++ b/frontend/apps/widget/src/components/preChatFormSchema.js
@@ -87,7 +87,17 @@ export const createPreChatFormSchema = (t, fields = []) => {
           })
         }
       }
-      
+
+      if (field.pattern) {
+        try {
+          fieldSchema = fieldSchema.regex(new RegExp(field.pattern), {
+            message: field.pattern_message || t('validation.invalid')
+          })
+        } catch {
+          // Ignore an invalid regex from the inbox config, field keeps its existing validation.
+        }
+      }
+
       if (field.required && field.type !== 'checkbox') {
         fieldSchema = fieldSchema.min(1, {
           message: t('globals.messages.required')
--- a/frontend/apps/widget/src/components/preChatFormSchema.test.js
+++ b/frontend/apps/widget/src/components/preChatFormSchema.test.js
@@ -76,6 +76,24 @@ describe('Pre Chat Text Fields', () => {
         expect(() => schema.parse({ thing: '' })).toThrow()
         expect(() => schema.parse({ thing: 'ok' })).not.toThrow()
     })
+
+    test('pattern rejects a value that does not match', () => {
+        const schema = build(field({ key: 'order', type: 'text', required: true, pattern: '^\\d{8}$', pattern_message: 'must be 8 digits' }))
+        expect(() => schema.parse({ order: '12345' })).toThrow('must be 8 digits')
+        expect(() => schema.parse({ order: '12345678' })).not.toThrow()
+    })
+
+    test('pattern is skipped for an optional empty value', () => {
+        const schema = build(field({ key: 'order', type: 'text', pattern: '^\\d{8}$' }))
+        expect(() => schema.parse({ order: '' })).not.toThrow()
+        expect(() => schema.parse({})).not.toThrow()
+        expect(() => schema.parse({ order: '123' })).toThrow()
+    })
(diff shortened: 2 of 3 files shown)


pujitha24:
Just checking in — this has been green and rebase-free for a week now; happy to adjust anything that would make it easier to review.

PR #568, opened 2026-09-05
skip away and offline agents in round-robin auto-assignment pool

### LINKED ISSUE #561: [Possible Bug] Don't assign chats to agents that are away
**Version:**
 - libredesk: 2.8
 
Auto assign still assigns chats if agents are away inactive (rather than marking themselves as away) 

Online              = "online"
Offline             = "offline"
Away                = "away"                  // Away due to inactivity
AwayManual          = "away_manual"           // Away due to manual setting from sidebar
AwayAndReassigning  = "away_and_reassigning"

// Skip user if availability status is `away_manual` or `away_and_reassigning`
if user.AvailabilityStatus == umodels.AwayManual || user.AvailabilityStatus == umodels.AwayAndReassigning {
    continue
}

Should be: 

if user.AvailabilityStatus == umodels.AwayManual ||
   user.AvailabilityStatus == umodels.AwayAndReassigning ||
   user.AvailabilityStatus == umodels.Away ||
   user.AvailabilityStatus == umodels.Offline {

pujitha24:
Motivation:
populateTeamBalancer() only excluded users with availability_status
`away_manual` or `away_and_reassigning` from the round-robin
auto-assignment pool. Idle `away` and `offline` agents were still
added to the pool, so conversations could be auto-assigned to agents
who are not actually online.

Approach:
Invert the check to an allowlist: only agents with availability_status
`online` are eligible for auto-assignment. This excludes `away`,
`away_manual`, `away_and_reassigning` and `offline` in one check, and
is also safer against any future/unknown status values.

Validation:
- go test ./internal/autoassigner/... passes with the fix
- The added test (TestPopulateTeamBalancer_OnlyOnlineUsersAddedToPool)
  fails against the pre-fix code (away/offline users leak into the
  pool) and passes after the fix, demonstrating this closes the
  reported gap
- go build ./..., go vet ./... and golangci-lint run on the changed
  package all pass clean
- go test -count=1 ./... has one unrelated pre-existing failure
  (TestNormalizeTimezoneFallsBackToUTC in internal/stringutil, an
  environment timezone-database casing issue) that reproduces
  identically without this change

Report: https://github.com/abhinavxd/libredesk/issues/561
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>
Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #561

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic assignment eligibility so only online users are included in the round-robin assignment pool.
  * Away, offline, and away-and-reassigning users are now excluded from automatic assignments.

* **Tests**
  * Added coverage verifying availability-based assignment eligibility.


abhinavxd:
Going offline due to inactivity or a lost connection doesn't mean the agent wants to stop accepting conversations, especially for email support. I'd keep the current behavior by default.

We could add an optional setting to stop automatic assignment after a configurable period of inactivity.
Something like what Intercom offers this through Automatic away mode, which is disabled by default: https://www.intercom.com/help/en/articles/6899988-automatic-away-mode
 
We could keep it under general settings

pujitha24:
Fair point, I hadn't thought about the case where an agent goes offline from a lost connection or inactivity but still wants email conversations queued up for them. I don't want to guess at the settings design though — should I revert this to the original behavior (only skipping `away_manual`/`away_and_reassigning`) and close this out, or would you rather this PR turn into the opt-in auto-away setting you described, under general settings with a configurable idle timeout? Happy to pick either up, just don't want to build the wrong one.

abhinavxd:
yes let’s keep the default and add an automatic away feature similar to Intercom. That would be really helpful!

pujitha24:
Sounds good. This branch is already back to matching `main` (the revert in the last commit just restores the original `away_manual`/`away_and_reassigning`-only skip, nothing else), so there's no further code change needed here for that part.

The automatic-away feature is bigger than I want to guess at in this PR — it'll need a new opt-in setting (disabled by default, configurable idle timeout) under general settings, plus whatever backend piece detects and applies the idle timeout. I'd rather scope that properly in its own issue/PR than bolt it onto this one. Want me to open a new issue for it, or would you prefer to write up the design yourself first?

chriswhitingBCH:
This should be for live chats only... They don't want to accept live chat requests when offline.

pujitha24:
@chriswhitingBCH that's a fair distinction, and it actually matches how #561 was originally worded ("don't assign chats to agents that are away") rather than all conversations. If/when the automatic-away setting @abhinavxd described gets built, scoping the idle-timeout behavior to live chat only (leaving email queued for offline agents) makes sense to me and is worth calling out explicitly in that design.

No code change on this PR though — as noted above, it's already back to matching `main`'s default behavior, and the new opt-in feature needs its own issue/PR rather than being guessed at here. @abhinavxd, let me know if you'd like me to open that follow-up issue (I can fold this live-chat-only scoping into it) or if you'd rather write up the design first.

chriswhitingBCH:
Thanks, I thought there was already an auto away function; sure I've seen myself go orange after locking my PC and then coming back to it. Whats the current behaviour for away/live chat allocation. Would the user have to manually set away or is there any other option we can look at?

chriswhitingBCH:
Another potential issue, useIdleDetection.js watches mousemove/keypress/click on the Libredesk page only, and after 5 min sets away. If they have the tab open in the background, then they get set away and dont get allocated chats when they are actually active using another tab/application.

Issue #574, opened 2026-09-08
Subjects for live-chat conversations: accept on widget init, expose on the widget API, editable by agents


pmslava:
## Use case

We run Libredesk as the support desk behind a mobile/web app. The app has an in-app "Support" screen that talks to the live-chat widget API through our own backend (which signs the identity JWT and exchanges it at `auth/exchange`). Users file titled requests, and agents triage the list in Libredesk by subject, exactly as they do for email conversations.

## Gap

- `POST /api/v1/widget/chat/conversations/init` only takes `message`, so a live-chat conversation is always created with an empty `subject`, although `CreateConversation` already accepts one and the conversation list already renders `conversation.subject` when it is set.
- The widget list/thread responses (`GET /api/v1/widget/chat/conversations`, `GET …/conversations/{uuid}`) do not expose `subject`, so a client cannot show it either.
- No route lets an agent set or correct a subject after creation, for any channel.

## Proposal (two small PRs)

1. **Widget: optional `subject` on init, returned on list/thread.** `chatInitReq` gains an optional `subject` (whitespace-collapsed to one line, max 255, 400 above); `handleChatInit` passes it to `CreateConversation` (no reference-number suffix, as for live chat today); `ChatConversation` gains `subject` and both widget queries select `COALESCE(c.subject, '')`. The embedded widget keeps sending no subject, so nothing changes for it. ~12 lines.
2. **Agents can edit a subject.** `PUT /api/v1/conversations/{uuid}/subject` with `{ "subject": "…" }` (same normalisation, empty clears), behind a new `conversations:update_subject` permission granted to Admin by migration (the same shape as `messages:write_private`), broadcasting the property change over the websocket, plus an inline edit of the subject in the conversation header for users holding the permission.

Happy to adjust naming or scope to fit the project's direction. PR for (1) follows; (2) is being prepared on top of it.

PR #575, opened 2026-09-08
Widget: accept an optional subject on conversation init and return it on the widget list and thread

pmslava:
Part 1 of #574.

Live-chat conversations are always created with an empty `subject`: the widget init request has no field for it, although `CreateConversation` already takes one and the conversation list already shows a subject whenever it is set. Clients that file titled requests through the widget API (an in-app support screen, in our case) therefore cannot give agents a subject to triage by, and cannot read one back either.

- `POST /api/v1/widget/chat/conversations/init` accepts an optional `subject`: whitespace-collapsed to a single line, max 255 characters (400 above). It is passed to `CreateConversation` as is, with no reference-number suffix, matching live chat today. The embedded widget keeps sending no subject, so its behaviour is unchanged.
- `GET /api/v1/widget/chat/conversations` and `GET …/conversations/{uuid}` now include `subject` (`""` when unset) via `COALESCE(c.subject, '')` in both widget queries and a `Subject` field on `ChatConversation`.

Tested with `go build ./...` and `go vet`, and against our own client: requests filed with a subject show it in the conversation list next to email tickets, and the widget list returns it. Part 2 (agents editing a subject, behind a new permission) follows as a separate PR.

## Summary by CodeRabbit

* **New Features**
  * Live chat conversations can include an optional subject.
  * Conversation subjects are normalized for consistent spacing and included in conversation data.
  * Subjects up to 255 characters are accepted.

* **Bug Fixes**
  * Subject length validation now counts characters correctly, including non-ASCII text.
  * Subjects exceeding 255 characters continue to be rejected with a bad-request response.

### DIFF
--- a/cmd/chat.go
+++ b/cmd/chat.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/abhinavxd/libredesk/internal/attachment"
 	bhmodels "github.com/abhinavxd/libredesk/internal/business_hours/models"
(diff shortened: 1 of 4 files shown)


pmslava:
Addressed the byte-vs-character finding in 8e62253: the limit is now counted with utf8.RuneCountInString after whitespace normalization, in a small helper with ASCII and non-ASCII boundary tests.

PR #576, opened 2026-09-08
Conversation: let agents set or change a conversation's subject from the desk

pmslava:
Part 2 of #574, stacked on #575 (the first two commits here are that PR; this PR's own change is the last commit, and it will be rebased once #575 lands).

Nothing in Libredesk could set or correct a conversation's subject after creation, for any channel. With subjects now arriving from the widget (#575), agents also need to fix or add one while triaging, the same way they change status or priority.

- `PUT /api/v1/conversations/{uuid}/subject` with `{ "subject": "…" }`: whitespace-collapsed to one line, max 255 characters (shared helper with the widget path), empty clears the subject (stored as NULL). Guarded by `enforceConversationAccess` like its status/priority siblings.
- New permission `conversations:update_subject`, added to the permission catalogue, the role editor, and `schema.sql` for fresh installs; the v2.9.0 migration grants it to every role that already holds `conversations:update_status` (Admin and Agent by default), following the `messages:write_private` precedent.
- The update records an activity entry ("<agent> changed the subject to …" / "cleared the subject") and broadcasts the property change to open desk tabs and the widget, so lists refresh live.
- UI: in the conversation header the subject (or "Add subject") becomes an inline input on click for users holding the permission; Enter/blur saves, Escape cancels. Shown at `md` and up; on phone widths the subject stays visible in the sidebar.

Not added: an automation event for subject changes (the automation catalogue has no such event today).

Tested: `go build`, `go vet`, `go test` for cmd, conversation, authz and migrations (the two migration tests need a live Postgres and skip without one); frontend `pnpm build:main` and `pnpm test:run` (782 tests) pass; `pnpm lint` reports no new errors on the touched files (the repo has pre-existing lint errors on `main`).


abhinavxd:
Hey,

This doesn't need a new permission as "conversation:write" is enough, Closing this.
This can be a simple `pencil` icon on sidebar where users with permissions change it and that's it.

PR #578, opened 2026-09-08
Conversation list: show a priority marker on each row

pmslava:
The conversation list shows no priority today, so an agent cannot see at a glance which rows are High priority (in our desk an automation sets High for paying customers, and the list is where triage happens).

- New `PriorityDot.vue`: an 8 px dot in the row's metadata cluster, left of the channel icon and the time, with a tooltip and `aria-label` "Priority: <name>". Colours use the design-system tokens already used by `SlaBadge.vue` (`bg-destructive` for High, `bg-warning-600` for Medium, `bg-muted-foreground` for Low or an unknown name); nothing renders when the conversation has no priority.
- `ConversationListItem.vue` mounts it. No backend change: the list query already selects the priority name and priority updates already reach list rows over the websocket.

Small and single-scope; tested with `pnpm build:main`, `pnpm test:run` (782 tests) and eslint on the touched files.

## Summary by CodeRabbit

- **New Features**
  - Added priority indicators next to conversation contact names.
  - Added tooltips showing localized priority labels.
  - Added visual priority icons with levels from unassigned to high priority.
  - Unrecognized priority values display an unknown marker.

### DIFF
--- a/frontend/apps/main/src/features/conversation/PriorityDot.vue
+++ b/frontend/apps/main/src/features/conversation/PriorityDot.vue
@@ -1,0 +1,44 @@
+<template>
+  <Tooltip v-if="priorityName">
+    <TooltipTrigger asChild>
+      <span
+        class="h-2 w-2 rounded-full flex-shrink-0"
+        :class="dotClass"
+        role="img"
+        :aria-label="priorityLabel"
+      />
+    </TooltipTrigger>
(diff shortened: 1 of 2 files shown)


abhinavxd:
Thank you, I made it a ladder instead of a dot as color blind folks won't be able to see it.

Issue #579, opened 2026-09-09
Custom attributes managed by an integration: mark definitions read-only for agents


pmslava:
## Use case

Contact custom attributes can be fed by an integration: the live-chat identity JWT's `contact_custom_attributes` are saved on the contact at every `auth/exchange`, and an external system (in our case the product's backend) keeps them current: plan, platforms, usage counters and similar.

## Gap

Every custom attribute is editable and deletable by any agent in the conversation sidebar. An agent can overwrite an integration-fed value by hand, the sidebar then shows something the source system never said, and the next exchange silently reverts it. There is no way to tell managed values from hand-maintained ones.

## Proposal

A per-definition flag `read_only` ("Managed by an integration"): the admin form gets a checkbox and the table a badge; the sidebar shows a lock instead of the pencil and trash for those attributes; the two manual update routes (`PUT /api/v1/conversations/{uuid}/custom-attributes` and `…/contacts/custom-attributes`) reject a change or removal of a managed key with a 400 naming it. Writes from the widget JWT, the pre-chat form and automations are unaffected. One `ALTER TABLE … ADD COLUMN IF NOT EXISTS read_only BOOLEAN NOT NULL DEFAULT false`. PR follows.

PR #580, opened 2026-09-09
Custom attributes: a per-definition "managed by an integration" flag that agents cannot edit

pmslava:
Implements #579.

Custom attributes fed by an integration (the live-chat identity JWT's `contact_custom_attributes`, refreshed on every `auth/exchange`) could be overwritten or deleted by any agent in the sidebar, and the next exchange silently reverted the edit. This adds a per-definition flag so an admin can mark such attributes as managed.

- `custom_attribute_definitions.read_only` (boolean, default false): schema, model, queries, and an idempotent `ALTER TABLE … ADD COLUMN IF NOT EXISTS` in the v2.9.0 migration with a test.
- Admin → Custom attributes: a "Managed by an integration (agents cannot edit the value)" checkbox and a "Managed" badge column.
- Conversation sidebar: a managed attribute shows its value with a lock and a tooltip instead of the pencil and trash; the inline editor cannot be opened for it.
- `PUT /api/v1/conversations/{uuid}/custom-attributes` and `…/contacts/custom-attributes` reject a change or removal of a managed key with a 400 (`errors.readOnlyCustomAttribute`), compared against the stored value; the widget JWT, pre-chat form and automation writes are untouched.

Tested: `go build`, `go vet`, `go test` for cmd, custom_attribute and migrations (migration tests run against the project's `make test-db` Postgres); `pnpm build:main` and `pnpm test:run` (785 tests, 3 new for the form schema); eslint on the touched files adds no errors.

## Summary by CodeRabbit

* **New Features**
  * Added support for marking custom attributes as managed and read-only.
  * Admins can configure this setting and identify managed attributes with a lock icon and “Managed” badge.
  * Read-only attributes remain visible but cannot be manually edited or removed.
* **Bug Fixes**
  * Prevented manual updates from changing or deleting values managed by integrations.
* **Tests**
  * Added coverage for read-only attribute validation and database migration behavior.


pmslava:
Addressed in c580130a: the checkbox has an id and the label references it, so the label text toggles it and screen readers get a name.

abhinavxd:
hey thanks I will check this over this week and add it to the next release.

This is a much needed feature.

abhinavxd:
I'll fix the merge conflict you can leave that for me.

Issue #581, opened 2026-09-09
Outbound reply guard: warn before sending internal text (mentions, hand-off headings, placeholders, internal words) to a customer


pmslava:
## Problem

A reply to the customer goes out with Ctrl+Enter or one click, and it is easy to send text that was meant for the team: an `@agent` mention that belongs in a private note, a "Draft reply:" heading an AI assistant wrote for the agent, an internal host or system name, a leftover `[placeholder]`. Nothing warns before the message leaves.

## Proposal

An outbound reply guard in the composer, reply mode only (private notes are exempt): before sending, the text is checked against a small rule set — `@mentions` that are not e-mail addresses, hand-off headings at line start ("Draft reply", "Proposed action", "Question for", "Why:", "Private note", "Internal"), placeholders (`[name]`, `{{ … }}`, `<insert`, TODO/FIXME/TBD/XXX, lorem ipsum) and an internal-vocabulary word list. A match opens a dialog listing what was found, with "Don't send" focused by default (Enter and Escape never send) and "Send anyway" as the only way through. The rules live in one module with a `{ id, label, test }` shape so the vocabulary can become an admin setting later; the first version ships a built-in list. PR follows.

PR #582, opened 2026-09-09
Conversation: guard a reply against internal text before it is sent

pmslava:
Implements #581.

A reply leaves with Ctrl+Enter or one click, so internal text meant for the team (an `@agent` mention, a "Draft reply:" heading written by an assistant, an internal system name, a leftover placeholder) can reach the customer with no warning. This adds an outbound reply guard, reply mode only; private notes are exempt.

- `replyGuard.js`: a pure rule set, each rule `{ id, label, test(text, context) → matches }`: `@mentions` that are not e-mail addresses (editor mention nodes included), hand-off headings at line start ("Draft reply", "Proposed action", "Question for", "Why:", "Private note", "Internal"), placeholders (`[name]`, `{{ … }}`, `<insert`, TODO, FIXME, TBD, XXX, lorem ipsum), and a built-in internal-vocabulary list matched whole-word, longest term first. 28 tests.
- `ReplyGuardDialog.vue` on the shared alert dialog: lists the matches grouped by rule; "Don't send" is the cancel element and receives focus on open, so Enter and Escape never send; only the "Send anyway" click proceeds.
- `ReplyBox.vue`: the guard runs in `processSend` before the other pre-send prompts, for both the Send button and Ctrl+Enter; everything else in the send path is unchanged.

The vocabulary is a single constant for now and is structured to become an admin setting. Tested with `pnpm build:main`, `pnpm test:run` (810 tests) and eslint on the touched files; not exercised in a running instance by me, only by the owner's own desk after this lands there.

## Summary by CodeRabbit

* **New Features**
  * Added a pre-send warning when a reply appears to contain internal notes, handoff instructions, mentions, or unfinished placeholders.
  * The warning lists detected items and lets agents cancel or send the reply anyway.
  * Improved detection for international words, characters, and mentions.
  * Added English text for the warning and its actions.

* **Tests**
  * Added coverage for detection, grouping, exclusions, international text, and message-type handling.


pmslava:
Addressed in 18e85bfd: the mention pattern and the whole-word boundaries now use Unicode letter, number and mark classes with the u flag, so @Мария is caught and josé@example.com is not, with tests for both.

pmslava:
Addressed in 0f16c381: the whole-word boundaries now include combining marks (\p{M}) like the mention pattern, with a test.

pmslava:
Addressed in 1d3ba64c: the test now uses "Slava" followed by U+0301, an actual combining mark, instead of the precomposed letter.

abhinavxd:
Closing this.

INTERNAL_VOCABULARY can't work. libredesk is used in many languages, and this list is your desk: pmslava, drifttt.com, Joplin, Paddle, RevenueCat.

Issue #585, opened 2026-09-09
Delete a conversation, optionally purging its source mails from the IMAP mailbox


pmslava:
## Problem

There is no way to delete a conversation. Spam and test conversations stay forever, and the only removal path, deleting the contact, takes every other conversation of that contact with it. For an email inbox even that is not enough: the source mails remain in the mailbox and the next scan re-imports them as a new conversation, because deduplication only sees the Message-IDs the desk still holds.

## Proposal

- `DELETE /api/v1/conversations/{uuid}` behind a new `conversations:delete` permission, granted to Admin by migration (like `contacts:delete`); it deletes the conversation with everything that belongs to it (messages, drafts, participants, tags, CSAT, SLA rows, AI events; attachment media rows and files explicitly, since media is polymorphic) in one transaction, and broadcasts a `conversation_delete` frame so open lists drop the row.
- For email inboxes, an optional mailbox purge (`purge_mail`, default true): each incoming message's Message-ID is searched in the inbox's IMAP mailbox with the channel's own connection settings, flagged `\Deleted` and expunged (`UID EXPUNGE` where supported). Message-IDs that could not be purged are returned so the UI can say so.
- UI: "Delete conversation" in the conversation's overflow menu for users holding the permission, with a confirmation dialog, a checkbox for the mailbox purge on email inboxes, and a toast naming any mails that survived.

PR follows.

PR #586, opened 2026-09-09
Conversation: delete a conversation, optionally purging its mails from the IMAP mailbox

pmslava:
Implements #585.

Conversations could not be deleted; the only removal was deleting the contact, and for email inboxes the source mails came back at the next scan because deduplication only sees Message-IDs the desk still holds.

- New permission `conversations:delete` (catalogue, role editor, `schema.sql` seed for Admin, and the v2.9.0 migration granting it to Admin idempotently, with a test).
- `DELETE /api/v1/conversations/{uuid}`: in one transaction it collects the incoming Message-IDs and the attachment media rows, deletes the media rows (polymorphic, so before the cascade removes the messages), then the conversation; `ON DELETE CASCADE` covers messages, drafts, participants, mentions, last-seen, tags, CSAT, SLA rows, AI and copilot rows and notifications (verified against a fixture). Afterwards the attachment files and thumbnails are removed through the media store, and a `conversation_delete` websocket frame reaches list subscribers.
- Optional mailbox purge for email inboxes (`purge_mail`, default true): the email channel's dial/auth code is shared with the scan (`dialIMAP`); for each Message-ID a UID SEARCH on the configured mailbox, `\Deleted` via UID STORE, then `UID EXPUNGE` when the server advertises UIDPLUS or IMAP4rev2, else `EXPUNGE`; unpurged Message-IDs are returned. The desk never appends sent mail to the mailbox, so only incoming messages are purged. Capped at two minutes; the desk-side deletion never fails because the purge did.
- UI: "Delete conversation" in the overflow menu (permission-gated), a destructive confirmation with the purge checkbox shown for email inboxes, navigation back to the list, and a toast that names surviving mails.

Tested: `go build`, `go vet`, `go test ./...` including the migration test against `make test-db`; the new SQL statements prepared and executed against a real Postgres with `schema.sql`; `pnpm build:main`, `pnpm test:run` (785 tests); eslint adds no errors. Not run against a live IMAP server by me.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to permanently delete conversations and their attachments.
  * Authorized users can optionally remove associated emails from connected mailboxes, moving them to Trash when available.
  * Added a conversation-deletion permission, enabled for administrators by default.
  * Conversation lists update automatically across active sessions after deletion.

* **Bug Fixes**
  * Deletion results now clearly indicate when emails could not be removed or were permanently expunged.
  * Improved email mailbox connection cleanup during failures.


pmslava:
Addressed in 6486104b: mails are now moved to the mailbox's Trash (discovered via SPECIAL-USE, with common names as a fallback) and expunged only when no Trash exists and UIDPLUS is available, so a plain EXPUNGE never runs; each UID SEARCH candidate is fetched and kept only when its Message-ID equals the stored one as a whole value; the response reports moved/expunged/not-found counts and any orphaned media files; the migration test now covers the upgrade path where Admin lacks the permission.

pmslava:
Addressed in d69a65bc: the media rows are no longer deleted in the transaction. Once their messages cascade away they are exactly what `get-unlinked-message-media` selects, so the periodic sweep retries any file the eager delete could not remove, and the thumbnail is deleted before the main file so a failed thumbnail keeps the row for the sweep too. The response field is renamed to `pending_media`. The no-Trash warning now states the outcome that follows from the UIDPLUS capability.

abhinavxd:
Will check it around next release (1-2 months) as its a pretty huge PR.

PR #589, opened 2026-09-09
Widget: keep manual contact edits when the JWT identity is synced

### LINKED ISSUE #588: Widget identity sync overwrites manual contact edits at every exchange
## Problem

`auth/exchange` syncs the contact's first name, last name, email and phone from the JWT whenever they differ from the contact, so an agent's manual correction of a contact (a real name instead of the e-mail local part an integration sends, a fixed typo) is overwritten at the customer's next visit. There is no way for a desk edit to win.

## Proposal

Remember on the contact what the integration last supplied (`users.external_sync`, jsonb) and overwrite a field only while the contact still holds that last-supplied value or is empty. A change at the source still reaches untouched fields; a value an agent changed by hand is kept; if the agent sets it back to the supplied value the sync resumes. Pure decision function with tests, one idempotent `ALTER TABLE users ADD COLUMN IF NOT EXISTS external_sync JSONB NOT NULL DEFAULT '{}'` in the unreleased migration, and the column is never serialized to the API. PR follows.

pmslava:
Fixes #588.

`resolveOrCreateExternalContact` synced first name, last name, email and phone from the JWT whenever they differed from the contact, so a manual correction by an agent was undone at the customer's next `auth/exchange`.

- `users.external_sync` (jsonb, default `{}`, never serialized to the API) remembers the values the integration last supplied; an idempotent `ADD COLUMN IF NOT EXISTS` sits in the v2.9.0 migration with a test.
- `ResolveExternalSync(current, synced, incoming)` decides per field: apply only when the contact is empty or still holds the last-supplied value; the remembered record follows the claim either way, so a field an agent sets back to the supplied value starts tracking again. Email is compared lower-cased and trimmed, as stored.
- `UpdateContactBasicInfo` takes the record as an optional parameter (nil for non-integration callers such as the AI agent tools, which leave it untouched); the external-id upsert seeds it on create and never wipes it on conflict.

Tested with `go build`, `go vet`, and `go test ./...` against `make test-db`, including the existing contact-sync integration tests; not exercised through a live `auth/exchange` by me.

## Summary by CodeRabbit

- **New Features**
  - Improved external contact synchronization to apply only externally managed identity updates while preserving manually edited contact details.
  - Contact email addresses are normalized by trimming whitespace and ignoring capitalization differences during synchronization.
  - Newly created and existing contacts retain the latest externally supplied identity values for more reliable future updates.
  - Added support for the `messages:write_private` permission for applicable roles.

- **Bug Fixes**
  - Prevented unchanged or manually corrected contact information from being unnecessarily overwritten during synchronization.


pmslava:
Addressed in c18b69ed. The sync now reads the contact with `FOR UPDATE`, decides and writes in one transaction (`SyncContactExternalIdentity` in the user manager), so an agent's edit or a second sync cannot land between the decision and the write; the widget handler only hands over the claim, and `UpdateContactBasicInfo` keeps its upstream signature. Email operands are compared lowercased and trimmed inside `ResolveExternalSync`, with cases for case and whitespace variants. The migration also seeds the record for contacts an integration created, since their current values are what it last supplied and an empty record would have frozen them on the next change.

PR #597, opened 2026-09-14
Relay WebSocket broadcasts across instances via a Redis backplane

ConradKurth:
## Problem

The WebSocket `Hub` (`internal/ws`) stores its client registry and conversation-subscription maps in process-local in-memory maps. Every broadcast path — `BroadcastMessage`, conversation-list subscriber pushes (`new_conversation` / `conversation_update` / `message_update` / `contact_update`), typing, and `KickUser` — only reaches clients connected to **the same instance**.

That means real-time updates silently break as soon as you run more than one instance behind a load balancer:

- An agent's browser holds its WebSocket on instance A.
- They reply to / resolve a conversation; that HTTP request is round-robined to instance B.
- Instance B mutates state and broadcasts on **its own** hub — instance A's socket never receives it.
- The inbox list doesn't update until the agent manually reloads (a fresh HTTP fetch reads current DB state).

So horizontal scaling / HA quietly disables live inbox refresh.

## Change

Add an optional cross-instance backplane, defaulting on via the existing Redis dependency (no new config):

- **`ws.Backplane`** interface (`Publish`/`Subscribe` of opaque payloads) with a Redis pub/sub implementation in **`internal/ws/redisbackplane`**.
- The hub tags each process with an instance ID, delivers every broadcast **locally as today**, and also publishes an envelope describing the delivery intent (users / conversations / kick + payload). Each instance replays envelopes from *other* instances against its own local registry and **skips its own** (deduped by origin) → exactly-once local delivery.
- **Graceful degradation:** with no backplane set, or if Redis is unavailable, delivery falls back to local-only. Publish failures are logged and never block the local broadcast, so a backplane outage degrades to single-instance behaviour rather than dropping events.
- Wired in `cmd/main.go` using the existing `*redis.Client` (already a hard dependency for auth / rate-limit / cache), so single-instance deployments are unaffected and multi-instance ones just work.

## Scope

Covers all agent-facing hub broadcasts: user-targeted messages (incl. notifications via the dispatcher), conversation-list subscriber updates, typing, and cross-instance `KickUser` (a revoked user is now disconnected on every instance).

Live-chat **widget** client delivery uses a separate per-instance registry (`internal/inbox/channel/livechat`) and has the same multi-instance limitation; I've left that for a follow-up to keep this PR focused. Happy to extend it here if you'd prefer.

## Tests

`internal/ws/backplane_test.go` covers: per-conversation dedup (a client subscribed to multiple conversations receives a broadcast once), remote-envelope delivery, own-envelope skip, and that the public fan-out methods publish. `go build ./...`, `go vet`, and existing `internal/ws` + `internal/conversation` tests pass; `gofmt` clean.

## Summary by CodeRabbit

- **New Features**
  - Added support for real-time WebSocket updates across multiple application instances.
  - Messages, conversation updates, typing indicators, and user session changes now synchronize between instances.
  - Added Redis-backed coordination for cross-instance WebSocket event delivery.

- **Bug Fixes**
  - Improved broadcast handling for users subscribed to multiple conversations, preventing duplicate updates.
  - Conversation broadcasts are now processed consistently, including when local subscribers are not present.


ConradKurth:
Thanks for the review. Addressing both:

**1. `Hub.publish` context (`backplane.go`)** — the current head already bounds it with `context.WithTimeout(context.Background(), publishTimeout)` (5s). I've left it as a plain deadline rather than also flipping `ContextTimeoutEnabled` on the shared app-wide `*redis.Client`, because that option changes timeout semantics for every other Redis user in the app (auth, rate-limit, cache, sessions) and go-redis already bounds a stuck `PUBLISH` via its default `WriteTimeout`. So the publish can't hang indefinitely regardless. Happy to enable it globally if you'd prefer that be the app default.

**2. Dropping under backpressure (`redisbackplane.go`)** — fixed in 0e6f32a. The receive loop now applies backpressure (blocks on delivery, with a ctx-cancel escape) instead of dropping, so control envelopes like `KickUser` aren't silently discarded, and the go-redis pub/sub channel is sized to match the buffer. The Hub's consumer delivers to local clients without blocking, so this only slows the reader under sustained overload rather than losing messages.

ConradKurth:
Good catch — fixed in 3bc64dc. Switched the forwarding loop from `PubSub.Channel` to `pubsub.ReceiveMessage(ctx)`, so a lagging consumer backpressures the Redis socket read directly instead of hitting `Channel`'s buffered-goroutine send timeout (which silently dropped after 60s). Added a small backoff on transient receive errors, and close `pubsub` when the loop stops.

Noted in the doc comment that Redis pub/sub is still non-durable across a subscriber disconnect (envelopes published during a reconnect gap aren't redelivered); a durable transport like Redis Streams would be the follow-up if control envelopes must survive disconnects. Kept that out of scope here to stay minimal.

abhinavxd:
Curious what's the need of having multi-instance libredesk?

Asking because libredesk is built single-instance today (Redis is cache/session only, not a coordination bus), and we run it at a fairly large scale on one process without hitting limits. Want to understand the use case before deciding on this.

ConradKurth:
@abhinavxd In my experiences it is always prudent to have any critical services be able to run in multiples in case one crashes, runs OOM or something like that. It is better to have redundancy in your system. Almost every system I have ever built will need to have at least 2-3 instances deployed in a production setting (except in very particular instances)

abhinavxd:
Supervisor can restart the process if it OOMs, that's a job scheduling problem.

I have a instance that gets 5000 tickets a day with 700 agents using it live. No issues.

Closing this as it isn't needed now.

abhinavxd:
Later if needed we can add this as of now, there's no need for this

PR #605, opened 2026-09-14
harden: cap request body on public CSAT submit endpoints

### LINKED ISSUE #604: [Hardening] Cap request body on the public CSAT submit endpoints
### Context
The public, unauthenticated CSAT submit endpoints — `POST /csat/{uuid}` and `POST /api/v1/csat/{uuid}/response` — accept a request body up to the global `max_body_size` (100MB) and buffer it before the handler truncates feedback to `maxCsatFeedbackLength`. A legitimate submission is tiny (rating + ≤1000-char feedback ≈ a few KB), so there's a large gap between what's accepted and what's ever needed.

Most deployments front the app with a reverse proxy that caps body size, which covers this. The gap only matters for instances run **without** such a proxy, where the 100MB global limit is the only bound on these public endpoints.

### Scope (being upfront)
This is modest, defense-in-depth hardening, not a high-severity issue. Because the body is already buffered by the server before the handler runs, an app-level cap rejects oversized submissions early but does **not** by itself prevent the initial buffering — a size-limiting proxy (or a lower global `max_body_size`) is still the way to stop that. The value here is bounding what the public handlers will process on a no-proxy deployment, with a clear 413.

### Proposed change
A small `limitBody(handler, maxBytes)` middleware (mirroring the existing `rateLimit` wrapper) applied to the two public CSAT submit routes, with a **24KB** cap — roughly 2× the largest legitimate submission (a 1000-char feedback is ~12KB worst-case when URL-encoded), so it never clips a real review while rejecting oversized bodies with `413`.

Happy to open a PR — I have it implemented with a unit test and a passing build. Wanted to check the direction fits first, per CONTRIBUTING.

anvme:
Closes #604

The public CSAT submit endpoints (`POST /csat/{uuid}`, `POST /api/v1/csat/{uuid}/response`) processed request bodies up to the global `max_body_size` (100MB) before truncating feedback. This adds a small `limitBody` middleware (mirroring `rateLimit`) and applies a **24KB** cap — ~2x the largest legitimate submission (a 1000-char feedback is ~12KB worst-case URL-encoded), so a real review is never clipped while oversized bodies get a `413`.

Note (also in #604): the body is already buffered by the server before the handler runs, so this rejects oversized submissions early but does not itself prevent the initial buffering — a size-limiting proxy or a lower global `max_body_size` is still needed for that. This bounds what the public handlers process on a no-proxy deployment.

Includes a unit test (`TestLimitBody`: under/at/over-limit). `go build ./...`, `go vet`, and the test pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - CSAT response submissions now reject request bodies larger than 24 KiB.
  - Oversized requests return a clear “Request Entity Too Large” response without being processed.
  - The size limit applies consistently to both API and widget CSAT response endpoints.


abhinavxd:
This isn't needed I think because the server already limits the maximum body size, in addition to our existing 1000 character limit

PR #613, opened 2026-09-17
Fix immediate attachment upload in widget chat 

csr4422:
## Summary #418 

- Queue attachments locally and upload them only when sending a message.
- Preview and remove queued files before sending.
- Prevent new attachments while a message is being sent.

## Summary by CodeRabbit

- **New Features**
  - Added a pending stage for file attachments before upload.
  - Selected or dropped files now appear immediately in the attachment preview.
  - Pending attachments can be removed before sending.
  - Messages can be sent with pending files, which upload automatically first.
- **Bug Fixes**
  - Attachment counts and send validation now include pending files.
  - Attachments cannot be added while a message is being sent.
  - Pending files are cleared when switching conversations.


abhinavxd:
incorrect fix, fix was needed for widget not agent side. Closing

